PR #9, opened 2026-01-02
fix: use python3 explicitly in patch-deps script

visionik:
## Summary
This PR fixes the patch-deps.sh script to explicitly detect and use python3, which is the standard on modern macOS systems.

## Changes
- Add python3/python detection logic with fallback
- Fail gracefully with error message if neither is found
- Add pnpm-lock.yaml to .gitignore

## Testing
Tested on macOS where only python3 is available by default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - docs: update build/test/release instructions
 - fix: stage message attachments in Messages attachments directory (thanks @antons)
 - feat: surface audio message transcriptions (thanks @antons)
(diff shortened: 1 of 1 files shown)


petter-b:
This looks like the most robust solution of the three PRs addressing this issue (#9, #10, #11). The explicit if/elif/else with a clear error message is nice.

One minor suggestion: consider splitting out the unrelated `.gitignore` change for `pnpm-lock.yaml` into a separate PR to keep this focused on the python fix.

PR #10, opened 2026-01-02
fix: use python3 instead of python in patch-deps.sh

darinkishore:
## Summary
- macOS no longer ships with `python` command, only `python3`
- Build fails on fresh macOS systems with `python: command not found`

## Test plan
- [x] Verified build completes successfully after change


petter-b:
This fixes the immediate issue, but PR #9 has a more robust solution with fallback to `python` for older systems and an explicit error message if neither is found. Might be worth closing in favor of #9.

PR #48, opened 2026-02-12
fix: reduce memory usage for large --limit values in chats command

### LINKED ISSUE #47: imsg chats gets SIGKILL'd with --limit 100 (works fine with --limit 30)
## Description

`imsg chats --limit 100` (with or without `--json`) consistently gets killed by SIGKILL before producing output. `imsg chats --limit 30` works fine every time.

## Environment

- **imsg version:** 0.4.0
- **macOS:** Darwin 25.2.0 (arm64, Apple Silicon)
- **Messages.app:** Signed in, Full Disk Access granted
- **Chat database:** ~4000+ message threads

## Steps to Reproduce

1. `imsg chats --limit 30` → works, output appears
2. `imsg chats --limit 100` → process receives SIGKILL, no output
3. `imsg chats --limit 100 --json` → same result, SIGKILL

## Expected Behavior

`imsg chats --limit 100` should return 100 chats.

## Actual Behavior

Process is killed by SIGKILL before producing any output. No error message or stderr output. Seems like it may be hitting a memory ceiling or triggering macOS jetsam during the larger query.

0xRaini:
## Problem

`imsg chats --limit 100` gets SIGKILL'd by macOS jetsam on databases with thousands of chats and millions of messages, while `--limit 30` works fine (#47).

## Root Cause

The `listChats()` SQL query used `JOIN chat_message_join × message` with `GROUP BY`, which forces SQLite to materialize all intermediate join rows into a temp B-tree before grouping. On large databases this creates a massive in-memory working set that triggers jetsam.

## Fix

Replace the JOIN + GROUP BY with a **correlated subquery** that computes `MAX(m.date)` per chat independently:

```sql
-- Before: materializes entire cross-product
SELECT ... FROM chat c
JOIN chat_message_join cmj ON c.ROWID = cmj.chat_id
JOIN message m ON m.ROWID = cmj.message_id
GROUP BY c.ROWID
ORDER BY last_date DESC LIMIT ?

-- After: computes max date per chat without cross-product
SELECT ...,
       (SELECT MAX(m.date)
        FROM chat_message_join cmj
        JOIN message m ON m.ROWID = cmj.message_id
        WHERE cmj.chat_id = c.ROWID) AS last_date
FROM chat c
WHERE last_date IS NOT NULL
ORDER BY last_date DESC LIMIT ?
```

Also adds `reserveCapacity` to the results array to reduce reallocation overhead.

## Testing

- All 89 existing tests pass
- Build succeeds on macOS (Apple Silicon)

Fixes #47

### DIFF
--- a/Sources/IMsgCore/MessageStore.swift
+++ b/Sources/IMsgCore/MessageStore.swift
@@ -88,18 +88,29 @@ public final class MessageStore: @unchecked Sendable {
   }
 
   public func listChats(limit: Int) throws -> [Chat] {
+    // Use a correlated subquery instead of JOIN + GROUP BY to avoid
+    // materializing millions of intermediate join rows in memory.
+    // The original query joined chat × chat_message_join × message and
+    // grouped the entire result, which caused SQLite to build a massive
+    // temp B-tree that could spike memory enough for macOS jetsam to
+    // SIGKILL the process (see #47).
     let sql = """
-      SELECT c.ROWID, IFNULL(c.display_name, c.chat_identifier) AS name, c.chat_identifier, c.service_name,
-             MAX(m.date) AS last_date
+      SELECT c.ROWID,
+             IFNULL(c.display_name, c.chat_identifier) AS name,
+             c.chat_identifier,
+             c.service_name,
+             (SELECT MAX(m.date)
+              FROM chat_message_join cmj
+              JOIN message m ON m.ROWID = cmj.message_id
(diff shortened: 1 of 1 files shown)


0xRaini:
@steipete CI green ✅ — replaces JOIN+GROUP BY with correlated subquery to prevent jetsam SIGKILL on large --limit values. Ready for review 🙏

PR #59, opened 2026-02-17
feat: IMCore bridge via dylib injection for typing indicators

### LINKED ISSUE #58: typing: IMChatRegistry never populates — entitlement blocker + any;-; GUID mismatch
## Summary

`imsg typing` is broken by two stacked bugs. Fixing either one alone won't make typing work. This consolidates #51, #54, and #56 with root cause analysis.

## Bug 1: GUID prefix mismatch (`any;-;` vs `iMessage;-;`)

When using `--to`, `ChatTargetResolver.directTypingIdentifier()` constructs `iMessage;-;+number` but the actual chat.db GUID is `any;-;+number` on modern macOS (Sequoia, Apple Silicon).

```sql
SELECT ROWID, guid, chat_identifier FROM chat;
-- 1|any;-;+12025551212|+12025551212
```

The `--chat-id` path works correctly (reads the GUID from DB and uses it as-is), but the `--to` path hardcodes the wrong prefix.

**Fix:** `directTypingIdentifier()` should read the actual GUID from chat.db for the recipient instead of constructing one. Or try all known prefixes (`any;-;`, `iMessage;-;`, `SMS;-;`).

## Bug 2: IMChatRegistry never populates (entitlement blocker)

This is the real blocker. Even with the correct GUID, `existingChatWithGUID:` and `existingChatWithChatIdentifier:` return nil because IMCore's chat registry is empty.

**What we tested (macOS 15.3, SIP disabled):**

1. `connectToDaemon` succeeds but `isConnected` stays `false`
2. `connectToDaemonWithLaunch:capabilities:blockUntilConnected:` — crashes with `NSProxy forwardInvocation:` (the daemon controller is a remote proxy that can't forward the call)
3. `blockUntilConnected` — same NSProxy crash
4. Retry loop with `RunLoop.current.run()` pumping for 10+ seconds — registry never populates
5. Ad-hoc codesigning with private entitlements (`com.apple.private.imcore.imhandle`, `com.apple.private.ids.messaging`) — AMFI kills the process with SIGKILL even with SIP disabled

**Root cause:** `IMChatRegistry` populates via XPC messages from `imagent`. The XPC connection requires Apple-internal entitlements that third-party binaries cannot claim without a valid Apple provisioning profile. AMFI enforces this even with SIP disabled.

**Evidence:** `amfid` log confirms the kill:
```
amfid: Restricted entitlements not validated, bailing out. 
Error Domain=AppleMobileFileIntegrityError Code=-420 "The signature on the file is invalid"
```

## Impact

Typing indicators are completely non-functional on `any;-;` GUID systems (which appears to be most/all modern macOS setups). Bug 1 is fixable. Bug 2 may require a fundamentally different approach:

- Injecting into Messages.app's process space (has the entitlements)
- Using XPC to talk to imagent through Messages.app
- AppleScript bridge (Messages.app scripting dictionary doesn't expose typing, but could potentially be extended)
- Xcode project with development signing (uncertain if Apple allows private entitlements even with dev profile)

## Environment

- imsg 0.5.0 (Homebrew)
- macOS 15.3 Sequoia, Apple Silicon (Mac Mini M4)
- SIP disabled via csrutil
- Messages.app running, imagent active
- All chats use `any;-;` GUID prefix

## Related

Consolidates #51, #54, #56

alexrudloffBot:
## Problem

`imsg typing` does not work because `IMChatRegistry` is empty when called from a standalone CLI process. Apple's entitlement wall blocks XPC connections to `imagent`. Documented in #58.

## Solution

Inject a helper dylib into Messages.app via `DYLD_INSERT_LIBRARIES`. Messages.app already has the required entitlements, so IMCore works fully inside its process. The CLI communicates with the injected dylib through file-based IPC (JSON files in the Messages container).

Same approach as [BlueBubbles](https://github.com/BlueBubblesApp/bluebubbles-helper) and [imsg-plus](https://github.com/micahbrich/imsg-plus).

### Architecture

```
CLI (imsg) <-> JSON files <-> Dylib (inside Messages.app) <-> IMCore/IMDaemon
```

**New files:**
- `Sources/IMsgHelper/IMsgInjected.m` — Obj-C injectable dylib (`__attribute__((constructor))`). Connects to IMDaemon, polls commands via NSTimer.
- `Sources/IMsgCore/MessagesLauncher.swift` — Kill/relaunch Messages.app with `DYLD_INSERT_LIBRARIES`.
- `Sources/IMsgCore/IMCoreBridge.swift` — Swift API wrapping file IPC.

**Modified:**
- `TypingIndicator.swift` — Prefers bridge, falls back to direct IMCore.
- `Makefile` — `build-dylib` target.

**New commands:** `imsg launch`, `imsg status`

### Also fixes
GUID prefix mismatch (#58): modern macOS uses `any;-;` prefix. The dylib tries all prefixes plus participant matching with phone normalization.

## Requirements
- SIP disabled
- Full Disk Access
- macOS 14+

## Tested
- `make build-dylib` + `swift build` — clean
- **Typing indicators confirmed working on recipient device**

Closes #58

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,15 +1,17 @@
 SHELL := /bin/bash
 
-.PHONY: help format lint test build imsg clean
+.PHONY: help format lint test build imsg clean build-dylib install
 
 help:
 	@printf "%s\n" \
-		"make format  - swift format in-place" \
-		"make lint    - swift format lint + swiftlint" \
-		"make test    - sync version, patch deps, run swift test" \
-		"make build   - universal release build into bin/" \
-		"make imsg    - clean rebuild + run debug binary (ARGS=...)" \
-		"make clean   - swift package clean"
+		"make format     - swift format in-place" \
+		"make lint       - swift format lint + swiftlint" \
+		"make test       - sync version, patch deps, run swift test" \
+		"make build      - universal release build into bin/" \
+		"make build-dylib - build injectable dylib for Messages.app" \
+		"make imsg       - clean rebuild + run debug binary (ARGS=...)" \
+		"make install    - build release binary and install to /usr/local/bin" \
+		"make clean      - swift package clean"
 
 format:
 	swift format --in-place --recursive Sources Tests
@@ -24,13 +26,26 @@ test:
 	scripts/patch-deps.sh
 	swift test
 
-build:
+build: build-dylib
 	scripts/generate-version.sh
 	swift package resolve
 	scripts/patch-deps.sh
 	scripts/build-universal.sh
 
-imsg:
+# Build injectable dylib for Messages.app (DYLD_INSERT_LIBRARIES).
+# Uses arm64e architecture to match Messages.app on Apple Silicon.
+# Requires SIP disabled on the target machine to inject into system apps.
+build-dylib:
+	@echo "Building imsg-bridge-helper.dylib (injectable)..."
+	@mkdir -p .build/release
+	@clang -dynamiclib -arch arm64e -fobjc-arc \
+		-Wno-arc-performSelector-leaks \
+		-framework Foundation \
+		-o .build/release/imsg-bridge-helper.dylib \
+		Sources/IMsgHelper/IMsgInjected.m
+	@echo "Built .build/release/imsg-bridge-helper.dylib"
+
+imsg: build-dylib
 	scripts/generate-version.sh
 	swift package resolve
 	scripts/patch-deps.sh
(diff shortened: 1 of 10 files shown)


steipete:
Closing as superseded by current main.

The main branch already contains the IMCore bridge / `imsg launch` / `imsg status` architecture from this direction, with additional safety around SIP status. I also landed f8f0c5d, which fixes the remaining prefix mismatch by normalizing `iMessage`, `SMS`, and `any` chat identifiers in both the Swift fallback and injected bridge lookup.

Proof on main: `make lint`, `make test`, `make build-dylib`, `make build`, plus green CI for f8f0c5d.

The macOS 26 library-validation / private-entitlement limitation called out in review is still real enough to track separately, so #60 remains open as the canonical issue for that platform blocker. Thanks for pushing the bridge approach forward.

PR #64, opened 2026-02-25
fix: deduplicate URL balloon messages in watch stream

lesaai:
## Problem

iMessage writes multiple rows to `chat.db` for the same URL when the rich link preview resolves. Both rows have `balloon_bundle_id` set to `com.apple.messages.URLBalloonProvider` with identical text and sender but different ROWIDs and timestamps (typically 30-65 seconds apart).

This causes downstream consumers (like OpenClaw) to process the same URL message twice, burning tokens and producing duplicate responses.

### Evidence from chat.db

```
4288|https://x.com/jakemclain_|com.apple.messages.URLBalloonProvider|793745478010970112
4290|https://x.com/jakemclain_|com.apple.messages.URLBalloonProvider|793745543638053888
```

Same sender, same text, same balloon type — two separate ROWIDs.

## Fix

In `messagesAfter()`, add `m.balloon_bundle_id` to the SELECT query and deduplicate URL balloon messages with the same `(sender, text)` within a batch. The first occurrence is kept; subsequent duplicates are skipped.

## Impact

- Only affects `messagesAfter()` (used by `watch` command)
- Only deduplicates messages with `balloon_bundle_id = 'com.apple.messages.URLBalloonProvider'`
- No impact on non-URL messages
- Build passes cleanly

### DIFF
--- a/Sources/IMsgCore/MessageStore+Messages.swift
+++ b/Sources/IMsgCore/MessageStore+Messages.swift
@@ -185,7 +185,8 @@ extension MessageStore {
              \(guidColumn) AS guid, \(associatedGuidColumn) AS associated_guid, \(associatedTypeColumn) AS associated_type,
              (SELECT COUNT(*) FROM message_attachment_join maj WHERE maj.message_id = m.ROWID) AS attachments,
              \(bodyColumn) AS body,
-             \(threadOriginatorColumn) AS thread_originator_guid
+             \(threadOriginatorColumn) AS thread_originator_guid,
+             m.balloon_bundle_id
       FROM message m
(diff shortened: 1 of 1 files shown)


steipete:
Landed via temp rebase onto main.\n\n- Gate: make lint && make test\n- Land commit: c9fa1c2ae27ea0464d0d2c72ab07c6aa4a369570\n- Merge commit: a2c16b3caccb144fa07d6cee5d2b0c20c582219a\n\nFollow-up hardening landed directly on main because push access to the fork head branch was denied (403).\n\nThanks @lesaai!

Issue #57, opened 2026-02-16
Bug: `is_group` field missing from JSON output


clawbunny:
## Summary

When using `imsg history --json` or `imsg watch --json`, messages from group chats do not include an `is_group` field. This makes it impossible for consumers to distinguish group messages from direct messages based on the JSON output alone.

## Current Behavior

```json
{
  "id": 1234,
  "chat_id": 4,
  "sender": "user@example.com",
  "text": "Hello",
  "is_from_me": false,
  "guid": "ABC-123",
  "created_at": "2026-02-16T12:00:00.000Z",
  "attachments": [],
  "reactions": []
}
```

No `is_group` field is present, regardless of whether the message is from a group chat or a DM.

## Expected Behavior

Messages should include `is_group: true` for group chats and `is_group: false` for DMs:

```json
{
  "id": 1234,
  "chat_id": 4,
  "sender": "user@example.com",
  "text": "Hello",
  "is_from_me": false,
  "is_group": true,
  "guid": "ABC-123",
  "created_at": "2026-02-16T12:00:00.000Z",
  "attachments": [],
  "reactions": []
}
```

## Root Cause

The macOS Messages SQLite database (`chat.db`) stores chat type in the `chat.style` column:
- **Style 43** = group chat (multi-party)
- **Style 45** = direct message (1:1)

This information is available via a JOIN between the `message`, `chat_message_join`, and `chat` tables, but `imsg` does not currently expose it in the JSON output.

## Suggested Fix

When building the JSON output for a message, JOIN against the `chat` table and set `is_group` based on `chat.style`:

```sql
SELECT 
  m.*,
  CASE WHEN c.style = 43 THEN 1 ELSE 0 END AS is_group
FROM message m
JOIN chat_message_join cmj ON m.ROWID = cmj.message_id
JOIN chat c ON cmj.chat_id = c.ROWID
```

Then include `"is_group": true/false` in the JSON output.

## Impact

Downstream consumers (automation tools, bots, AI gateways) that process iMessage via `imsg` cannot reliably determine if a message came from a group chat. They must resort to workarounds like maintaining a separate config-based allowlist of known group chat IDs.

## Environment

- imsg version: 0.4.0
- macOS: 15.x (Sequoia)
- Architecture: arm64 (Apple Silicon)

steipete:
Fixed on `main` in 6b08ee1.

[...]

PR #70, opened 2026-03-02
feat: Add isGroup support to detect group chats

PeterRosdahl:
## Summary
Adds the ability to detect whether a message is from a group chat.

## Changes
- Add `isGroup` boolean to `Message` struct
- Update SQL queries to JOIN against `chat` table and fetch `chat.style`
- Map `chat.style == 43` → `isGroup = true`  
- Add `is_group` field to JSON output

## Usage
```bash
imsg history --json | jq '.messages[0].is_group'
```

Useful for filtering out group chat messages or handling them differently.

PR #71, opened 2026-03-02
feat: Add isGroup support to detect group chats

PeterRosdahl:
## Summary
Adds the ability to detect whether a message is from a group chat.

## Changes
- Add `isGroup` boolean to `Message` struct
- Update SQL queries to JOIN against `chat` table and fetch `chat.style`
- Map `chat.style == 43` → `isGroup = true`
- Add `is_group` field to JSON output

## Testing
Tested with synthetic data by copying chat.db and modifying `style` to verify both cases:
- `style=45` → `is_group: false` ✅
- `style=43` → `is_group: true` ✅

## Usage
```bash
imsg history --chat-id 3 --json | jq '.is_group'
```

---

*Note: Apologies for PR #70 which was closed - I was a bit too quick on the keyboard and submitted before proper testing. This one has been verified.*

### DIFF
--- a/Sources/IMsgCore/MessageStore+Messages.swift
+++ b/Sources/IMsgCore/MessageStore+Messages.swift
@@ -51,6 +51,8 @@ extension MessageStore {
     let associatedTypeColumn = hasReactionColumns ? "m.associated_message_type" : "NULL"
     let destinationCallerColumn = hasDestinationCallerID ? "m.destination_caller_id" : "NULL"
     let audioMessageColumn = hasAudioMessageColumn ? "m.is_audio_message" : "0"
+    let chatStyleColumn = hasChatStyleColumn ? "c.style" : "NULL"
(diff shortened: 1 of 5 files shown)


steipete:
Thanks @PeterRosdahl. This is now covered on `main` by 6b08ee1, which fixed #57 by routing `imsg history --json` and `imsg watch --json` through the same chat-aware payload path as RPC.

Current CLI JSON now includes `chat_identifier`, `chat_guid`, `chat_name`, `participants`, and `is_group`, with regression coverage for both history and watch. I went with the existing chat metadata helper instead of adding `Message.isGroup` to the core model here, so this PR is superseded rather than something we should merge as-is.

Verified in 6b08ee1 with:
- `swift test --filter 'historyCommandRunsWithChatID|watchCommandRunsWithJsonOutput|watchCommandFlushesJsonOutput'`
- `make lint`
- `make test`

steipete:
Superseded by the main fix in 6b08ee1; closing to keep the queue clean. Thanks again for the tested repro and implementation direction.

PR #81, opened 2026-04-05
fix: Slow JSON history output

kacy:
## Summary

This speeds up `imsg history --json` by preloading attachments and reactions in bulk instead of doing per-message metadata lookups while streaming the response.

## What Changed

- preload attachment metadata for the selected history rows in one query
- preload reaction metadata once and map it back to the target messages in memory
- keep reaction handling consistent with the existing single-message path, including reactions whose rows are missing a `chat_message_join` entry
- move the history metadata helpers into a dedicated `MessageStore+HistoryMetadata.swift` extension so lint stays clean

## Why

`history --json` was much slower than plain history output because the JSON path fetched attachments and reactions for each message individually. The reaction lookup was the main cost, and it repeated a scan-heavy query for every message in the result set.

## Impact

For a local run against chat `1` with `--limit 200`:

```bash
/usr/bin/time -p ./.build/debug/imsg history --chat-id 1 --limit 200
real 2.29
user 0.53
sys 0.20

/usr/bin/time -p ./.build/debug/imsg history --chat-id 1 --limit 200 --json
real 2.75
user 0.69
sys 0.27
```

Before the change, the same JSON command was taking about 55 seconds locally.

## Root Cause

The JSON history path did N+1 metadata fetches. In practice, the repeated reaction lookups dominated runtime because each message triggered another reaction query.

## Verification

- `make lint`
- `make test`


steipete:
Applied a narrower replacement on `main` in `9d92837`.

What I kept from this PR:
- batch attachment metadata for the selected history rows
- batch reaction metadata for JSON history output
- keep the existing single-message reaction path for watch/RPC payloads
- add regression coverage for both batch attachment and batch reaction mapping

One important adjustment: I did not take the unbounded reaction preload as-is. The landed version batches reactions only for the selected history message GUIDs, so `history --json` does not scan and map unrelated reaction rows from the whole database.

Verified locally:
- `swift test --filter 'attachmentsByMessagesReturnsMetadataByMessageID|bulkReactionsForMessagesGroupsByMessageID|bulkReactionsReturnsEmptyWhenColumnsMissing|historyCommandRunsWithChatID'`
- `make lint`
- `make test` (135 tests)
- `make build`

Thanks @kacy for tracking down the JSON history N+1 path and the clear benchmark.

PR #87, opened 2026-04-11
fix: re-confirm is_from_me after INSERT to prevent echo loops

clawdcode:
## Problem

When `imsg send` sends a message, macOS writes `is_from_me=0` on the initial SQLite INSERT, then updates it to `1` shortly after. The filesystem watcher fires on the INSERT event and emits the message with `is_from_me: false`, causing agents (e.g. OpenClaw) to treat their own sent messages as inbound — triggering an infinite echo loop.

## Fix

When a new message arrives with `is_from_me=false` and no `handle_id` (the fingerprint of a potentially-unconfirmed outbound message), delay emission by `reconfirmDelayMs` (default: 400ms) and re-read the row from the DB. If `is_from_me` is now `true`, suppress the event. If still `false`, emit normally.

## Changes

- `MessageWatcherConfiguration`: new `reconfirmDelayMs` param (default `400`, set `0` to disable and preserve original behavior)
- `MessageWatcher/WatchState`: re-confirmation logic with a `pendingReconfirm` set to avoid duplicate scheduling
- `MessageStore+Messages`: new `message(rowID:)` single-row lookup
- `WatchCommand`: new `--reconfirm-delay` flag (same duration format as `--debounce`, default `400ms`)
- `RPCServer+Handlers`: new optional `reconfirm_delay_ms` param in `watch.subscribe`

## Backwards compatibility

Setting `reconfirmDelayMs = 0` / `--reconfirm-delay 0` preserves the exact original behavior with zero overhead.

### DIFF
--- a/Sources/IMsgCore/MessageStore+Messages.swift
+++ b/Sources/IMsgCore/MessageStore+Messages.swift
@@ -42,6 +42,89 @@ extension MessageStore {
     return try messages(chatID: chatID, limit: limit, filter: nil)
   }
 
+  /// Fetch a single message by its ROWID. Returns nil if not found.
+  public func message(rowID: Int64) throws -> Message? {
+    let bodyColumn = hasAttributedBody ? "m.attributedBody" : "NULL"
+    let guidColumn = hasReactionColumns ? "m.guid" : "NULL"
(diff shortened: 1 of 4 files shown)


clawdcode:
Closing — will test the fix locally first before submitting upstream.

PR #88, opened 2026-04-17
feat: CLI JSON group metadata + imsg group command

mryanb:
## Summary

Brings group-chat metadata parity between the direct CLI (`imsg chats`, `imsg history`, `imsg watch`) and the existing JSON-RPC surface, and adds a focused `imsg group --chat-id N [--json]` command for looking up a chat's identity + participants. All additions are optional JSON fields — no existing keys are renamed or removed.

## Motivation

The RPC path (`imsg rpc`) already returns `is_group`, `chat_identifier`, `chat_guid`, `chat_name`, and `participants` via helpers in `RPCPayloads.swift`. The direct CLI `--json` output does not. Downstream consumers that parse `imsg chats --json` or stream from `imsg watch --json` currently can't distinguish groups from 1:1s, pull a group's display name, or enumerate participants without switching to RPC. This change closes that gap using the existing `MessageStore.chatInfo` / `participants` helpers.

### `imsg chats --json`

Adds `guid`, `display_name`, `is_group`, `participants`.

```json
{"id":1234,"display_name":"Family","guid":"any;+;chat1234567890123456789","name":"Family","last_message_at":"2026-01-15T10:00:00.000Z","service":"SMS","is_group":true,"participants":["+16125551100","+16125551101","+16125551102","+16125551103"],"identifier":"chat1234567890123456789"}
```

### `imsg history --json` and `imsg watch --json`

Adds `chat_identifier`, `chat_guid`, `chat_name`, `participants`, `is_group` on every message payload.

```json
{"sender":"+16125551111","is_from_me":true,"text":"...","chat_name":"","chat_guid":"any;+;abc123def456abc123def456abc123de","chat_id":42,"is_group":true,"chat_identifier":"abc123def456abc123def456abc123de","participants":["+16125552222","jimbo@example.com"],"destination_caller_id":"+16125551111", ...}
```

### `imsg group --chat-id <N> [--json]`

New read-only subcommand that prints id, identifier, guid, name, service, is_group, and participants. Works for direct chats too (`is_group: false`).

```
id: 42
identifier: abc123def456abc123def456abc123de
guid: any;+;abc123def456abc123def456abc123de
name:
service: iMessage
is_group: true
participants:
  - +16125552222
  - jimbo@example.com
```

### Plain-text output

Minimal group indicators, chosen to avoid breaking parsers that key off the existing `^\d{4}-` timestamp prefix:

- `imsg chats` — group rows render as `[<id> group]` instead of `[<id>]`.
- `imsg history` — a single header line `# group chat: <name> (participants: ...)` prints before the messages when the chat is a group. Per-message line format unchanged.
- `imsg watch` — appends ` (group)` to the end of the per-message line. Timestamp prefix unchanged.

## Internals

- `isGroupHandle(identifier:guid:)` moved into `IMsgCore` (new `GroupDetection.swift`) so the CLI and RPC paths share one detection helper. Semantics unchanged (`;+;` only; `;-;` is not a group, matching the existing test expectation).
- `actor ChatCache` extracted from `RPCServer+Support.swift` into its own file and reused by `imsg watch` to amortize per-chat `chatInfo` / `participants` lookups across a stream. RPC behavior unchanged.
- `Chat` gained optional `guid` and `displayName` via defaulted initializer params — additive, source-compatible.
- `MessageStore` detects `chat` and `chat_handle_join` table presence at init (new `hasChatTable` / `hasChatHandleJoinTable` flags, same pattern as the existing `hasReactionColumns` flag) and gates `chatInfo` / `participants` on them. Keeps minimal-schema test fixtures working.

## Notes worth a second look

- **`docs/groups.md` correction**: the prior doc stated that `SERVICE;-;TARGET` also marks a group. That contradicts the authoritative test at `Tests/imsgTests/RPCPayloadsTests.swift:10` which asserts `iMessage;-;chat999` is **not** a group, and it's inconsistent with Messages.app's own convention (`-` = direct, `+` = group). Updated the doc to match the code.
- **`participants` excludes the local user**, because `chat_handle_join` only stores external handles. This is inherited from Messages.app's schema, not new behavior — but it was never documented. Added an explicit note in `docs/groups.md` and the CHANGELOG entry so downstream consumers don't build rosters under a wrong assumption. `destination_caller_id` (already exposed per message) is the signal for which local handle routes the chat.
- **New public IMsgCore surface**: `isGroupHandle(identifier:guid:)` is now a public top-level function in IMsgCore. Library consumers gain one symbol.

## Non-breaking

- No existing JSON keys renamed or removed.
- Nil optionals are omitted by Swift's default encoder, so the CLI JSON shape for consumers that don't care about group metadata is byte-identical to before when `chatInfo` isn't populated.
- RPC JSON output is unchanged — the RPC dictionary path at `RPCPayloads.swift:25` already overlays its own keys on top of `MessagePayload.asDictionary()`.
- `Chat`'s memberwise init still works positionally and by keyword, thanks to defaulted new params.

## Testing

- `make test`: **107 tests pass** (baseline on `main` was 97; this PR adds 10 new). No existing test removed or weakened.
- `swift format lint --recursive Sources Tests`: clean.
- `swiftlint` not run — not installed locally; happy to add an output from wherever the CI runner has it.
- End-to-end smoke on a live `~/Library/Messages/chat.db`:
  - `imsg chats --json` returned real groups across iMessage, SMS, and RCS with all new fields present.
  - `imsg group --chat-id N --json` on a real group — `GroupPayload` shape clean.
  - `imsg send --chat-id N --text "..."` to a real group — AppleScript resolved the GUID via the existing `ChatTargetResolver` path and delivered successfully.
  - `imsg history --chat-id N --limit 2 --json` on the same group — the just-sent message carried `is_from_me:true`, `is_group:true`, `chat_identifier`, `chat_guid`, `chat_name`, `participants`, and the existing `destination_caller_id`.

## Commits

Landed as 4 logical commits to keep review and bisect straightforward:

1. `refactor: move isGroupHandle into IMsgCore; extract ChatCache actor`
2. `feat: expose group metadata in CLI JSON for chats/history/watch`
3. `feat: add imsg group command`
4. `docs: document CLI group metadata; clarify participants excludes self`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Changelog
 
 ## Unreleased
+- feat: expose group-chat metadata (`is_group`, `chat_identifier`, `chat_guid`, `chat_name`, `participants`, raw `display_name`, `guid`) in `imsg chats`, `imsg history`, and `imsg watch` JSON output — matching the existing RPC surface. `participants` reports external handles only (the local user is implicit, as in the underlying `chat_handle_join` table).
+- feat: new `imsg group --chat-id <id> [--json]` command prints chat identity and participants for one chat id.
+- feat: `imsg chats` plain text marks groups with a `[<id> group]` tag; `imsg history` prints a group-chat header; `imsg watch` appends ` (group)` to group messages.
+- refactor: move `ChatCache` actor into its own `Sources/imsg/ChatCache.swift`; reuse it in `imsg watch` to memoize per-chat lookups.
+- refactor: move `isGroupHandle(...)` from the CLI target into `IMsgCore` so both CLI JSON and RPC paths share one group-detection source of truth.
 - fix: dedupe URL balloon preview duplicates in watch stream without cross-chat/schema regressions (#64, thanks @lesaai)
 - fix: remove non-functional `typing` command and related RPC methods
 - fix: remove unsupported standalone IMCore typing path and stale error branch
--- a/README.md
+++ b/README.md
@@ -24,7 +24,8 @@ make build
 ```
 
 ## Commands
-- `imsg chats [--limit 20] [--json]` — list recent conversations.
+- `imsg chats [--limit 20] [--json]` — list recent conversations (group chats are marked with `[group]` in plain text and carry `is_group`, `guid`, `display_name`, and `participants` in JSON).
+- `imsg group --chat-id <id> [--json]` — show identity and participants for one chat (works for direct chats too).
 - `imsg history --chat-id <id> [--limit 50] [--attachments] [--participants +15551234567,...] [--start 2025-01-01T00:00:00Z] [--end 2025-02-01T00:00:00Z] [--json]`
 - `imsg watch [--chat-id <id>] [--since-rowid <n>] [--debounce 250ms] [--attachments] [--participants …] [--start …] [--end …] [--json]`
 - `imsg send --to <handle> [--text "hi"] [--file /path/img.jpg] [--service imessage|sms|auto] [--region US]`
@@ -34,8 +35,11 @@ make build
 # list 5 chats
 imsg chats --limit 5
 
-# list chats as JSON
+# list chats as JSON (includes is_group, guid, display_name, participants)
 imsg chats --limit 5 --json
+
+# show identity + participants for one chat (works for direct chats too)
+imsg group --chat-id 1 --json
 
 # last 10 messages in chat 1 with attachments
 imsg history --chat-id 1 --limit 10 --attachments
--- a/Sources/IMsgCore/GroupDetection.swift
+++ b/Sources/IMsgCore/GroupDetection.swift
@@ -1,0 +1,11 @@
+import Foundation
+
+/// Returns `true` when the chat identifier or GUID encodes a group chat.
+///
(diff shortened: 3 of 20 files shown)


steipete:
Thanks @mryanb. I applied the still-missing pieces on `main` in 518144e, adjusted to the current code after #57 had already landed.

What landed:
- `imsg group --chat-id <id> [--json]`
- `imsg chats --json` now includes `guid`, `display_name`, `is_group`, and `participants`
- README + `docs/groups.md` updates, including the `;+;` group marker and participant semantics
- Changelog credit for #88

What I did not carry over:
- The broader `ChatCache` extraction / `IMsgCore` group-detection refactor, because current `main` already shares the message metadata path for history/watch and the remaining feature did not need that wider API change.

Verified:
- focused group/chats tests
- `make lint`
- `make test` (131 tests)
- `make build`
- black-box smoke with built `bin/imsg` for `chats --json`, `group --json`, `group --help`, plus universal binary/codesign checks.

steipete:
Applied the useful remaining pieces on `main` in 518144e; closing this PR as superseded/applied. Thanks again for the thorough implementation and live-smoke notes.

mryanb:
Hey @steipete thanks for taking in, looking forward to this release hitting Homebrew.

PR #89, opened 2026-04-19
feat: --resolve-replies flag inlines referenced message on history/watch

lesaai:
## Motivation

LLM agents running on top of `imsg watch` have a recurring failure mode: when a user sends a native iMessage reply (or a thread reply) the agent sees two blobs of text back-to-back — the quoted context and the new message — with no structural separator. When the quoted context happens to be the agent's own prior message, models read it as "my own message came back to me" and loop or re-send.

Downstream consumers work around this with bash wrappers that run a second `sqlite3` process against `chat.db` to look up `reply_to_guid` and prepend a synthetic `[Reply to #N: ...]` line. That adds a process per event, runs an unpooled connection, and only covers one of the two reply mechanisms (`reply_to_guid` misses `thread_originator_guid`, which is what most users actually hit on modern iOS/macOS).

This change lets `imsg` itself resolve the referenced message in the same connection that's already open, with a single bounded `SELECT`, and emit it as structured data.

## Changes

- **`MessageStore.messageByGUID(_:)`** — new public API on `IMsgCore`. Single bounded `SELECT ... WHERE m.guid = ? LIMIT 1`, mirrors the existing `messages()` decode path (attributedBody, audio transcription, sender fallback, reaction metadata). Returns `nil` for empty/whitespace/unknown GUIDs, doesn't throw.
- **`--resolve-replies` flag** on `imsg history` and `imsg watch`. Opt-in, so the baseline JSON shape is unchanged for consumers that don't pass it.
- **JSON mode** adds two optional nested fields:
  - `reply_to` — present when `reply_to_guid` resolves. Object: `{id, guid, sender, is_from_me, text, created_at}`
  - `thread_originator` — present when `thread_originator_guid` is distinct from `reply_to_guid` and resolves. Same shape.
  - Failed lookups (GUID not in DB) omit the nested field but still emit the outer message.
- **Text mode** prefixes a `  ↳ reply to <sender> #<id>: <excerpt>` line (or `↳ thread to …` when only `thread_originator_guid` is present) before the replying message. Excerpt is truncated to 80 chars, newlines collapsed.
- Pre-existing `reply_to_guid` and `thread_originator_guid` fields remain in place.

## Example

With a message that replies-in-thread to an earlier one:

**Baseline (unchanged):**
```json
{"id":7425,"reply_to_guid":null,"thread_originator_guid":"356418B7-...","sender":"+15551234567","text":"i didnt say Joerges…",...}
```

**With `--resolve-replies`:**
```json
{"id":7425,"reply_to_guid":null,"thread_originator_guid":"356418B7-...","thread_originator":{"id":7424,"guid":"356418B7-...","sender":"...","is_from_me":true,"text":"Yeah. And it's okay to not know...","created_at":"2026-04-18T18:15:55Z"},"sender":"+15551234567","text":"i didnt say Joerges…",...}
```

**Text mode with `--resolve-replies`:**
```
  ↳ thread to +15551234567 #7424: Yeah. And it's okay to not know. The whole architecture exist…
2026-04-18T18:17:01Z [recv] +15551234567: i didnt say Joerges…
```

## Tests

Added `Tests/IMsgCoreTests/MessageStoreByGUIDTests.swift` covering:
- empty / whitespace GUID → nil
- unknown GUID → nil
- exact-match resolution returns the decoded `Message`
- `LIMIT 1` behaviour when duplicate GUIDs exist (defensive — shouldn't happen in practice)

Full suite compiles fine on your CI with Xcode 16+ Swift Testing. Can't run it locally — my Mac has Command Line Tools only (Swift 6.2.4 but no Xcode), which is missing the Swift Testing framework. `main` has the same build issue locally so this isn't new. Flagging in case you want me to migrate the new tests to XCTest instead; happy to do that if preferred.

## Scope

- No changes to existing JSON schema for consumers that don't pass the flag.
- No changes to the schema-detection path (columns used already detected as `hasReactionColumns` / `hasThreadOriginatorGUIDColumn`).
- `--resolve-replies` costs ~1 extra SELECT per message that has a reply target, on the same pooled connection. On a normal watch stream (single-digit msgs/sec) the overhead is negligible.
- Flag is not added to the RPC surface; happy to extend `watch.subscribe` with `include_reply_context` in a follow-up if that's desired.

## Context

This unblocks LLM-driven agents using `imsg watch` as their input channel, specifically one project (OpenClaw's iMessage plugin) where the double-display issue was causing real user-facing confusion with messages that referenced prior agent output.

Happy to iterate on naming (`--resolve-replies` vs `--reply-context` vs `--inline-replies`), field names (`reply_to` vs `replied_to`), or the text-mode prefix character if you'd like something more compatible with non-UTF8 terminals.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## Unreleased
+- feat: `--resolve-replies` flag on `history` and `watch` resolves `reply_to_guid` and `thread_originator_guid` to the referenced message inline (thanks @lesaai). Adds optional nested `reply_to` / `thread_originator` objects in JSON output and `↳` prefix lines in text output. Opt-in so existing consumers see no schema change.
+- feat: `messageByGUID(_:)` on `MessageStore` for library consumers that need to resolve a GUID to a `Message` without a second process (thanks @lesaai).
 - fix: dedupe URL balloon preview duplicates in watch stream without cross-chat/schema regressions (#64, thanks @lesaai)
 - fix: remove non-functional `typing` command and related RPC methods
 - fix: remove unsupported standalone IMCore typing path and stale error branch
--- a/README.md
+++ b/README.md
@@ -25,8 +25,8 @@ make build
 
 ## Commands
 - `imsg chats [--limit 20] [--json]` — list recent conversations.
-- `imsg history --chat-id <id> [--limit 50] [--attachments] [--participants +15551234567,...] [--start 2025-01-01T00:00:00Z] [--end 2025-02-01T00:00:00Z] [--json]`
-- `imsg watch [--chat-id <id>] [--since-rowid <n>] [--debounce 250ms] [--attachments] [--participants …] [--start …] [--end …] [--json]`
+- `imsg history --chat-id <id> [--limit 50] [--attachments] [--resolve-replies] [--participants +15551234567,...] [--start 2025-01-01T00:00:00Z] [--end 2025-02-01T00:00:00Z] [--json]`
+- `imsg watch [--chat-id <id>] [--since-rowid <n>] [--debounce 250ms] [--attachments] [--resolve-replies] [--participants …] [--start …] [--end …] [--json]`
 - `imsg send --to <handle> [--text "hi"] [--file /path/img.jpg] [--service imessage|sms|auto] [--region US]`
 
 ### Quick samples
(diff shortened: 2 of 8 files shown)


lesaai:
Closing for now. I need to validate this change end-to-end against the OpenClaw iMessage pipeline on our fork first (there is a related internal bug report about reply-context handling that wasn't cross-checked before this PR went up). Will reopen after proper internal review + install + verification. Apologies for the noise.

PR #96, opened 2026-04-30
ci: update Homebrew tap on release

dinakars777:
Adds an `update-homebrew-tap` job that dispatches the centralized formula updater in `steipete/homebrew-tap` after a successful release. Uses `gh run watch` for synchronous failure visibility.

Depends on steipete/homebrew-tap#29

**Required secret:** `HOMEBREW_TAP_TOKEN` with workflow dispatch access to `steipete/homebrew-tap`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -100,3 +100,55 @@ jobs:
           fi
 
           gh release edit "$TAG" --notes-file "$notes_file"
+
+  update-homebrew-tap:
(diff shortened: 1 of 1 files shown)


dinakars777:
Closing this — same issue as steipete/remindctl#49. The primary release flow is local signing + notarization, so wiring the tap update into the CI workflow wouldn't cover real releases. Will follow up with a PR that integrates the tap updater into the local release scripts instead.

PR #99, opened 2026-05-05
Fix parseAttributedBody silently dropping messages longer than 127 bytes

SagarSDagdu:
## Background

The `attributedBody` column in `chat.db` is a serialized NSAttributedString (Apple typedstream format). `TypedStreamParser.parseAttributedBody` (`Sources/IMsgCore/TypedStreamParser.swift`) scans for two byte markers around the string segment:

- start marker: `0x01 0x2B`
- end marker: `0x86 0x84`

The string segment between those markers begins with a BER-style length prefix:

| First byte    | Meaning                                                  |
|---------------|----------------------------------------------------------|
| `0x00..0x7F`  | The first byte itself is the length (single-byte form).  |
| `0x81 NN`     | One length byte follows. Length = `NN`.                  |
| `0x82 NN NN`  | Two length bytes follow, big-endian.                     |

The current decoder only handled the single-byte form, with this heuristic:

```swift
if segment.count > 1, Int(segment[0]) == segment.count - 1 {
    segment.removeFirst()
}
```

Any message of length >= 128 uses `0x81` or `0x82` as the first prefix byte. The heuristic does not match, so the prefix is left in place. The bytes `0x81` and `0x82` are invalid UTF-8 leading bytes, so strict UTF-8 decoding fails (or, with the lossy `String(decoding:as:)` already in use, produces a U+FFFD replacement that pollutes the "longest segment" comparison) and the function silently emits `""`. Downstream callers then drop the message entirely.

## Concrete repro

A real attributedbody blob from a 140-byte SMS message starts with:

```
... 95 84 01 2B 81 8C 00 37 39 36 33 35 34 20 69 73 20 4F 6E 65 ... 86 84 02 49 ...
                ^^^^^^^^                                          ^^^^^^^^
                start marker, then:                                 end marker
                  81 = "1-byte length follows"
                  8C = 140 (decimal)
                  00 = NSStringEncoding tag (NUL, gets trimmed)
                  37 39 36 ... = "796..." (the actual text)
```

Segment after the start marker: `81 8C 00 37 39 36 ...`. The heuristic compares `0x81 (= 129)` against `segment.count - 1 (= 142)`. They do not match, so the prefix stays. UTF-8 decoding fails on `0x81`. Result is `""`.

## The fix

Replace the inline single-form prefix handling with a `decodeSegment` helper that tries every plausible prefix length (0, 1 if the original heuristic matches, 2 for `0x81`, 3 for `0x82`) and returns the longest *strictly* valid UTF-8 decoding. Strict decoding (via `String(bytes:encoding:.utf8)?`) is what makes the comparison work — otherwise the lossy variant always succeeds with replacement characters and the wrong prefix length wins.

Behavior summary:

- Short message (e.g. `[0x05, 'h','e','l','l','o']`): prefix-len 1 wins, same output as before.
- Medium message (length 128..255, prefix `0x81 NN`): prefix-len 2 strips `81 NN`, body decodes cleanly. Previously dropped.
- Long message (length > 255, prefix `0x82 NN NN`): prefix-len 3 strips `82 NN NN`, body decodes. Previously dropped.
- Garbage segments fall through to `0` (no prefix) and the existing whole-blob UTF-8 fallback.

Short messages still hit the original path, so this is fully backward compatible.

## Tests

Added four focused unit tests in `Tests/IMsgCoreTests/UtilityTests.swift` exercising `TypedStreamParser.parseAttributedBody` directly:

1. `typedStreamParserDecodesShortSingleBytePrefix` — regression guard for the original single-byte path.
2. `typedStreamParserDecodesMediumMessageWith0x81Prefix` — 140-byte ASCII body with `0x81 NN` prefix. Returns `""` without the fix.
3. `typedStreamParserDecodesLongMessageWith0x82Prefix` — 300-byte ASCII body with `0x82 NN NN` prefix. Returns `""` without the fix.
4. `typedStreamParserHandlesMixedBinaryNoise` — segment whose first byte matches neither form and is not `count - 1`. Asserts no crash and that the parser falls back to no-prefix decoding.

Full suite (`swift test`): 192 tests pass.

## Test plan

- [x] `swift test` — all 192 tests pass locally
- [x] New tests fail on `main` and pass on this branch
- [x] No public API changes
- [x] No new dependencies

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
 - fix: route default bridge calls over v2 IPC when available and reject
   unsupported `chat-create --service SMS` requests instead of reporting a
   service that was not applied.
+- fix: decode typedstream attributed bodies with `0x81`/`0x82` length prefixes
+  so long fallback message text is preserved in history and watch output (#99,
+  thanks @SagarSDagdu).
 
 ### Docs And CI
 - docs: publish the per-feature docs site at `imsg.sh` and add
--- a/Sources/IMsgCore/TypedStreamParser.swift
+++ b/Sources/IMsgCore/TypedStreamParser.swift
@@ -19,13 +19,8 @@ enum TypedStreamParser {
       if bytes[index] == start[0], bytes[index + 1] == start[1] {
         let sliceStart = index + 2
         if let sliceEnd = findSequence(end, in: bytes, from: sliceStart) {
-          var segment = Array(bytes[sliceStart..<sliceEnd])
-          // Check if first byte equals length prefix (convert byte to Int for comparison)
-          if segment.count > 1, Int(segment[0]) == segment.count - 1 {
-            segment.removeFirst()
-          }
-          let candidate = String(decoding: segment, as: UTF8.self)
-            .trimmingLeadingControlCharacters()
+          let segment = Array(bytes[sliceStart..<sliceEnd])
+          let candidate = decodeSegment(segment)
           if candidate.count > best.count {
             best = candidate
           }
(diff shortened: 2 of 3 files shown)


SagarSDagdu:
@steipete I encountered messages of length >= 128 in my own `chat.db`, hence implemented this fix. It works accurately with longer messages while preserving original implementation intact.
Please review.

steipete:
Thank you!

steipete:
Thanks @SagarSDagdu. This fix is already represented on `main` and shipped in `0.7.0` via the typedstream length-prefix parser coverage (`0x81` / `0x82` attributed-body decoding). The current `0.7.3` release includes it as well, and CI is green on the released branch.

Closing this PR as superseded by the maintainer-landed version so the queue reflects the shipped state.

PR #100, opened 2026-05-06
feat: port BlueBubbles private-API surface as imsg bridge (fixes #60)

omarshahine:
## Summary

Adds a manual rewrite/port of the BlueBubbles private-API surface as an `imsg` bridge, layered on top of the existing AppleScript send path. None of BlueBubbles' source or binaries are vendored — the API vocabulary (`send-rich`, `tapback`, `chat-mark`, etc.) is preserved so existing BB tooling can target this. The bridge is opt-in (SIP-off + injected helper dylib) and does not affect default `send`/`history`/`watch`/`rpc` flows.

**Fixes #60** — `imsg typing` now works on macOS 26 (Tahoe) when the bridge is active. The dylib runs in-process within Messages.app, so `setLocalUserIsTyping:` uses Messages.app's own entitlements and bypasses imagent's private-entitlement check. `TypingIndicator.setTyping` prefers the bridge path and only falls back to the direct IMCore path (the one that fails on Tahoe per #60) when the bridge is unavailable.

Provenance: manual port inspired by the BlueBubbles helper/server (Apache-2.0). The action verb vocabulary was preserved intentionally for tooling compatibility; the implementation is local and lives entirely in this repo. Upstream references:
- BlueBubbles Helper: https://github.com/BlueBubblesApp/bluebubbles-helper
- BlueBubbles Server: https://github.com/BlueBubblesApp/bluebubbles-server
- Text formatting reference: https://github.com/BlueBubblesApp/bluebubbles-helper/pull/50

`imsg` remains under MIT.

## What's new

**Bridge command surface** (require `imsg launch` + SIP off):
- Messaging: `send-rich`, `send-multipart`, `send-attachment`, `tapback`
- Mutation (macOS 13+, selector-gated via `imsg status`): `edit`, `unsend`, `delete-message`, `notify-anyways`
- Chat management: `chat-create`, `chat-name`, `chat-photo`, `chat-add-member`, `chat-remove-member`, `chat-leave`, `chat-delete`, `chat-mark`
- Introspection: `account`, `whois`, `nickname`, `search`
- Live events via `watch --bb-events` (typing, alias-removed)

**Text formatting on macOS 15+**: `--format` JSON ranges with bold/italic/underline/strikethrough on `send-rich`. Earlier macOS silently degrades to plain text (private `IMText*` attribute names don't exist before Sequoia).

**v2 IPC** to replace the v1 single-file polling (which raced under concurrent CLI invocations):

```
[...]

PR #101, opened 2026-05-06
fix: macOS 26 bridge regressions — typing/read RPC, effect mapping, attachment registration

omarshahine:
## What this fixes

PR #100 landed the BlueBubbles bridge (typing, send-rich, tapbacks, chat management, JSON-RPC). It passed tests, but on macOS 26 it didn't actually work end-to-end: `send-rich`, replies, tapbacks, typing, and read were silent no-ops against a live Messages.app. This PR makes the bridge functional on macOS 26 by aligning it with BlueBubblesHelper's verified reference implementation.

Three things were broken.

**1. Message construction silently dropped on macOS 26.** Every send path went through `+initIMMessageWith…:expressiveSendStyleID:`, which on macOS 26 returns a message whose `IMMessageItem` has empty `bodyData`. imagent reads `bodyData` (NSArchiver typedstream) when persisting, so empty-payload messages get dropped. `send-rich`, `send-multipart`, tapbacks, and threaded replies all looked fine in code and never reached chat.db.

The construction pipeline now follows BB-Helper's macOS 26 approach: allocate `IMMessageItem` via the 9-arg `initWithSender:time:body:attributes:fileTransferGUIDs:flags:error:guid:threadIdentifier:`, NSArchiver-archive the `attributedBody` and `setBodyData:` it onto the item, apply extended fields (effectId, subject, associated-message metadata) before wrapping (the `_imMessageItem` accessor returns a transient item rebuilt each call, so post-wrap setters don't persist), wrap with `+messageFromIMMessageItem:sender:subject:`, and dispatch via `-_sendMessage:adjustingSender:shouldQueue:`. The public `sendMessage:` silently no-ops on macOS 26 when the item has `sender = nil`. Replies derive `threadIdentifier` by loading the parent and calling the IMCore C function `IMCreateThreadIdentifierForMessagePartChatItem` (resolved via `dlsym`, since the symbol now lives only in the dyld shared cache). Reactions get their own `+instantMessageWithAssociatedMessageContent:…:threadIdentifier:` path with the `p:<part>/<parent-guid>` prefix matching BB-Helper. The legacy initializer remains as a fallback for older macOS.

**2. The JSON-RPC surface drifted from the bridge surface.** `typing` and `read` were registered in the bridge but never wired into JSON-RPC, so `imsg rpc` returned `methodNotFound`. `status --json` didn't advertise `rpc_methods`, so consumers (Lobster's openclaw imessage plugin) gated those indicators off entirely. Eight chat/group handlers (`chats.create`/`delete`/`markUnread`, `group.rename`/`setIcon`/`addParticipant`/`removeParticipant`/`leave`) didn't survive the #100 merge: they shipped as bridge selectors but were never exposed as JSON-RPC methods. All bridge methods are now wired through and the full list is advertised in `status --json` (15 total, was 7).

**3. Selector vocabulary drifted from BB-Helper.** Six bridge IMCore selectors didn't match the verified macOS 26 names (group ops, `markChatItemAsNotifyRecipient:`, `markLastMessageAsUnread`, block-based message lookup). Effect short-names (`invisibleink`) were sent as-is where iMessage requires bundle IDs (`com.apple.MobileSMS.expressivesend.invisibleink`). ARC retention on the IMD persistent path was loose. All replaced with BB-Helper's verified vocabulary.

A file-based debug logger was added alongside the fixes because macOS 26 unified-logging redacts NSLog from system-app processes, making these regressions invisible from the outside. It is load-bearing for diagnosing this class of issue.

## Verification

206 tests pass. `swift-format lint` clean. `swiftlint` clean on changed files (1 pre-existing file-length warning).

```
$ imsg send-rich --chat 'any;-;+15551234567' --text 'plain'
  → 245-byte attributedBody lands in chat.db
$ imsg send-rich --chat ... --text '...' --effect invisibleink
  → expressive_send_style_id = com.apple.MobileSMS.expressivesend.invisibleink
$ imsg send-rich --chat ... --text 'reply' --reply-to <parent-guid>
  → thread_originator_guid = parent
$ printf '...' | imsg rpc
  → typing/read return {"ok":true}; status --json advertises 15 rpc_methods
```

## Known gaps (not introduced by this PR)

- Tapback persistence to chat.db is finicky on a Messages session that has been force-restarted multiple times; clean on a stably-synced session. Receiver-visible behavior on iPhone still needs human verification.
- Attachment registration via `send-attachment` doesn't always link the row even with `prepareOutgoingTransfer` (suspected: IMFileTransferCenter requires Messages.app fully connected). The CLI `imsg send --file` (AppleScript path) is unaffected. Tracked in #103.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Private API Bridge
+- fix: restore macOS 26 bridge sends, replies, tapbacks, typing/read RPC, and
+  chat/group lifecycle RPC methods after the BlueBubbles-inspired bridge port
+  regressed on Tahoe (#101, thanks @omarshahine).
+
 ## 0.7.2 - 2026-05-06
 
 ### Release Packaging
--- a/Sources/IMsgHelper/IMsgInjected.m
+++ b/Sources/IMsgHelper/IMsgInjected.m
@@ -956,10 +956,19 @@ static id constructIMMessageViaItem(NSAttributedString *attributedText,
     NSArray *transferGuids = fileTransferGuids ?: @[];
     NSError *err = nil;
     NSString *guid = [[NSUUID UUID] UUIDString];
-    // 0x5 = IMMessageItemFlagsFromMe | IMMessageItemFlagsFinished. Audio
-    // messages historically use 0x300005; subject-bearing messages 0x10000d.
-    // Items with `flags = 0` are dropped before reaching chat.db on macOS 26.
-    unsigned long long flags = isAudioMessage ? 0x300005ULL : 0x5ULL;
+    // BlueBubblesHelper-verified flag set: 0x100005 (FromMe | Finished |
+    // 0x100000 finalize bit) for normal text+attachment, 0x10000d when a
+    // subject is set, 0x300005 for audio messages. The earlier `0x5`
+    // variant was the cause of malformed attachments on the receiver — the
+    // 0x100000 bit is what tells imagent to finalize the payload.
+    unsigned long long flags;
+    if (isAudioMessage) {
+        flags = 0x300005ULL;
+    } else if (subject.length) {
+        flags = 0x10000dULL;
+    } else {
+        flags = 0x100005ULL;
+    }
     id sender = nil;
     NSDictionary *attributes = nil;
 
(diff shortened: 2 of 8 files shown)


steipete:
Maintainer triage pass:

- GitHub CI is green on the PR head.
- I merged the PR head onto current `main` locally and ran `make lint`: clean except the existing `BridgeMessagingCommands.swift` file-length warning.
- `make test` passes locally on the merge result: 206 tests.
- `make build-dylib` also succeeds; only the pre-existing `NSKeyedUnarchiver.unarchiveObjectWithData` deprecation warning remains.

No blocking code issue from this pass. One maintainer blocker before landing: this is a user-facing private-API/RPC fix, so it needs an `Unreleased` changelog entry crediting @omarshahine. Runtime proof still relies on the Tahoe/macOS 26 smoke evidence in the PR description; I did not run a live Messages.app smoke on this machine.

steipete:
Thanks @omarshahine. The macOS 26 bridge regression fixes from this PR were landed on `main` in `2d7b506` / `3695bfb` follow-ups and are now shipped in `0.7.3`: https://github.com/openclaw/imsg/releases/tag/v0.7.3

Closing this PR as superseded by the released maintainer-landed version.

PR #105, opened 2026-05-06
fix(security): clamp IPC dirs to 0700 and reject symlinked paths

omarshahine:
## Summary

Hardens the IMCore bridge's IPC layer against same-UID exfiltration attacks. Three layers of mitigation, all at IPC boundaries:

- **0700 on the RPC queue dirs** (`.imsg-rpc/`, `.imsg-rpc/in/`, `.imsg-rpc/out/`). Mode is enforced both at `createDirectory:attributes:` time and via a final `chmod()` to fix dirs left 0755 by prior unsandboxed runs.
- **Symlink rejection at every component**, not just the final one. `SecurePath.hasSymlinkComponent` (Swift) + `pathHasSymlinkComponent` (Obj-C twin in the dylib) walk each component with `lstat()` and reject any `S_IFLNK`. Catches parent-dir symlinks that a `realpath()`-vs-lexical-input check would miss — macOS rewrites `/tmp` → `/private/tmp`, so that approach breaks on legitimate paths anyway.
- **Strict throws on queue dir creation/cleanup**, replacing `try?` swallowing in `MessagesLauncher`. The bridge now refuses to start on an insecure path instead of silently operating on it.

## Threat model

A same-UID attacker — another user process, a sandboxed peer with home-directory reach — could otherwise:
1. Drop a file in `.imsg-rpc/in/` (writable when the dir was 0755) and inject an arbitrary RPC.
2. Symlink `~/.ssh/id_rsa` (or a password-manager DB) into a path the bridge accepts as an attachment, and have Messages.app exfiltrate it to an attacker-controlled handle.

## Surface

| File | Change |
|---|---|
| `Sources/IMsgCore/SecurePath.swift` | New: lexical lstat walk |
| `Sources/IMsgCore/IMsgBridgeClient.swift` | `ensureDirectory` mkdir-with-0700 + pre/post symlink check |
| `Sources/IMsgCore/MessagesLauncher.swift` | `ensureSecureQueueDirectory` + throwing `cleanQueueDirectory` |
| `Sources/IMsgHelper/IMsgInjected.m` | `pathHasSymlinkComponent` + `ensureSecureDirectory` helpers; applied in `handleSendAttachment` and `startV2InboxWatcher` |
| `Tests/IMsgCoreTests/UtilityTests.swift` | Two unit tests covering final + parent-component symlink detection |

## Test plan

- [x] `swift build` clean
- [x] `make build-dylib` clean (only the pre-existing `unarchiveObjectWithData:` deprecation warning, unchanged from main)
- [ ] `swift test --filter securePath` — needs to be run on a host with a swift-testing-capable toolchain (Xcode); the maintainer machine had Command Line Tools only
- [ ] Smoke test on Tahoe: launch with bridge, confirm `.imsg-rpc/` is `drwx------`, confirm `imsg send-attachment --file <symlink>` errors with `"Symlinked attachment paths are not allowed"`, confirm normal attachment still works

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
   chat/group lifecycle RPC methods after the BlueBubbles-inspired bridge port
   regressed on Tahoe (#101, thanks @omarshahine).
 
+### Security
+- fix: harden bridge IPC queue directories and attachment paths against
+  symlink traversal while preserving trusted macOS system aliases like `/tmp`
+  (#105, thanks @omarshahine).
+
 ## 0.7.2 - 2026-05-06
 
 ### Release Packaging
--- a/Sources/IMsgCore/IMsgBridgeClient.swift
+++ b/Sources/IMsgCore/IMsgBridgeClient.swift
@@ -126,6 +126,9 @@ public final class IMsgBridgeClient: @unchecked Sendable {
   }
 
   private func ensureDirectory(_ path: String) throws {
+    if SecurePath.hasSymlinkComponent(path) {
+      throw IMsgBridgeError.ioError("\(path) traverses a symlink")
+    }
     var isDir: ObjCBool = false
     if FileManager.default.fileExists(atPath: path, isDirectory: &isDir) {
       if isDir.boolValue { return }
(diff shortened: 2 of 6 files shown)


steipete:
Draft CI triage: the failing tests are exposing a real path-semantics issue, not just inverted expectations.

`SecurePath.hasSymlinkComponent(target.path)` and `realChild.path` return `true` in CI because `FileManager.default.temporaryDirectory` on macOS traverses a system symlink such as `/var` -> `/private/var`. With the current lexical `lstat()` walk, legitimate `/tmp`/`/var`-based paths will be rejected too, which conflicts with the PR body note about macOS rewriting `/tmp` to `/private/tmp`.

So the fix should define the intended policy explicitly: either normalize trusted system aliases like `/tmp`/`/var` before walking, or keep the stricter policy and update docs/tests to say those paths are intentionally rejected. I would not just flip the expectations.

steipete:
Thanks @omarshahine. The IPC directory/symlink hardening from this PR was landed on `main` in `2432269` and is now shipped in `0.7.3`: https://github.com/openclaw/imsg/releases/tag/v0.7.3

Closing this PR as superseded by the released maintainer-landed version.

PR #107, opened 2026-05-07
fix: prefer structured length prefix over raw decode in typedstream

SagarSDagdu:
## Summary

- Found in further testing of #99: when a typedstream segment's body length falls in **32–126 bytes**, the unstripped `prefixLen = 0` decode beats the correct `prefixLen = 1` decode by one character, prepending the length byte (`'@'`, `' '`, `'~'`, `'-'`, `'*'`, etc.) to message text.
- `decodeSegment` now ranks structured BER-style prefix candidates (1, 2, or 3 bytes) ahead of the raw whole-segment decode. The raw decode is only used as a fallback when no structured prefix produces valid UTF-8.
- Switched `Set<Int>` to ordered `[Int]` so tie-breaking is deterministic.

## Where the old code failed

For a body of N bytes where `N ∈ [32, 126]` and N is itself a printable ASCII byte:
- `prefixLen = 0` → decodes whole segment to **N+1 chars** (length byte is printable, not trimmed).
- `prefixLen = 1` → decodes body to **N chars**.

The "longest valid" tiebreak picked the unstripped candidate. The 5/140/300-byte fixtures added in #99 didn't surface this because their length bytes were either control chars (trimmed) or invalid UTF-8 leads (failed entirely) — the failure window sits exactly in the gap between those buckets.

## Test plan

- [x] Verified the bug exists pre-fix: stashed source, ran new tests against the #99 code, all four failed with exactly the predicted output (e.g. `"@AAAA…A"` instead of `"AAAA…A"`).
- [x] All 11 `typedStreamParser*` tests pass with the fix applied.
- [x] New regression tests cover:
  - 32-byte body / length `0x20` (lower edge of window)
  - 64-byte body / length `0x40` (mid)
  - 126-byte body / length `0x7E` (upper edge — `0x7F` is DEL/control)
  - 48-byte multibyte UTF-8 body (12 × 🎉) / length `0x30` — proves the fix isn't ASCII-specific

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 - docs: document Linux as read-only support for existing copied Messages
   databases.
 
+### Message Decoding
+- fix: strip printable typedstream length bytes from recovered `attributedBody`
+  text for 32-126 byte messages (#107, thanks @SagarSDagdu).
+
 ## 0.7.3 - 2026-05-06
 
 ### Private API Bridge
--- a/Sources/IMsgCore/TypedStreamParser.swift
+++ b/Sources/IMsgCore/TypedStreamParser.swift
@@ -38,36 +38,43 @@ enum TypedStreamParser {
   }
 
   /// Strips a typedstream length prefix from `segment` and returns the longest valid UTF-8 decoding.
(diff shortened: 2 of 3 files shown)


SagarSDagdu:
@steipete Another fix hardening the stream parsing in case of short messages. Please review 🙏🏻

PR #119, opened 2026-05-20
Return send/create identity in RPC responses

svetly:
This PR fixes the core send/create identity surface.

- send returns chat_guid and message guid when observable
- send returns delivered service/transport when observable
- chats.create completes and returns created chat/message identity

Follow-up PR #120 exposes handles.check over RPC.

### DIFF
--- a/Sources/IMsgCore/MessageStore+Messages.swift
+++ b/Sources/IMsgCore/MessageStore+Messages.swift
@@ -286,45 +286,48 @@ extension MessageStore {
 
     return try withConnection { db in
       let rows = try db.prepareRowIterator(query.sql, bindings: query.bindings)
-      guard let row = try rows.failableNext() else { return nil }
-      let decoded = try decodeMessageRow(
-        row,
-        columns: query.selection.columns,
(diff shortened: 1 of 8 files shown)


steipete:
Thanks for the PR. I could not push fixups to the contributor branch even though GitHub reported maintainer edits were enabled, so I recreated the change locally with your author credit preserved.

Landed in #127 / f65a7fc15d43600b0b1a4a950445c9face0f126c.

Additional fixes included before landing:
- latest sent-message verification now preserves newest decoded-message ordering for repeated attributed-body sends
- explicit SMS chat creation now refuses non-SMS handle fallback instead of silently creating an iMessage chat
- changelog entry added with thanks

Proof on #127:
- macOS CI passed
- Linux read-core CI passed
- focused local Swift tests passed
- make lint passed with existing file-length warnings
- bridge helper dylib build passed with the existing Foundation deprecation warning
- autoreview clean

For future PRs, enabling and testing direct maintainer branch writes should avoid this recreate-and-close path.

PR #136, opened 2026-06-01
fix: fail open on undetermined contacts auth

### LINKED ISSUE #135: Bug: imsg chats/history can hang when Contacts auth is not determined
### Summary

On macOS 26.4.1, `imsg chats` and `imsg history` can hang indefinitely in CLI usage when Contacts permission has not yet been decided (`CNContactStore.authorizationStatus(for: .contacts) == .notDetermined`).

This appears to be caused by `ContactResolver.create()` awaiting `requestAccess(for: .contacts)` from a CLI path. In this environment, the callback never resolves promptly, so the command appears to hang.

### What works

- `imsg group --chat-id 1090 --json` returns immediately
- direct Hermes SQLite fallback works
- Messages DB is readable enough for non-Contacts-dependent flows

### What hangs

- `imsg chats --limit 1 --json`
- `imsg chats --limit 1`
- `imsg history --chat-id 1090 --limit 1 --json`

### Evidence

Direct Swift probe:

```bash
swift -e 'import Contacts; print(CNContactStore.authorizationStatus(for: .contacts).rawValue)'
```

returns:

```text
0
```

(`.notDetermined`)

And:

```bash
swift -e '
import Contacts
import Foundation
let s = CNContactStore()
print("before", CNContactStore.authorizationStatus(for: .contacts).rawValue)
let sem = DispatchSemaphore(value: 0)
s.requestAccess(for: .contacts) { granted, err in
  print("granted", granted)
  print(err?.localizedDescription ?? "nil")
  sem.signal()
}
let r = sem.wait(timeout: .now() + 5)
print("wait", r == .success ? "done" : "timeout")
'
```

outputs:

```text
before 0
wait timeout
```

So the permission request path itself is the hang.

### Likely root cause

`ChatsCommand.run()` and `HistoryCommand.run()` eagerly call:

```swift
await ContactResolver.create()
```

And `ContactResolver.create()` currently does:

```swift
case .notDetermined:
  let granted = await requestAccess(store: store)
  return granted ? load(...) : NoOpContactResolver(...)
```

That blocks core read commands on a permission flow that may not complete in CLI/headless contexts.

### Proposed fix

Fail open for `.notDetermined`:

```swift
case .notDetermined:
  return NoOpContactResolver(contactsUnavailable: true)
```

This preserves core `chats` / `history` behavior and only disables optional contact-name enrichment.

### Better long-term design

Make contact resolution opt-in (for example `--resolve-contacts`) instead of automatic for all `chats` / `history` calls.

cemendes:
## Summary

Avoid hanging `imsg chats` and `imsg history` when Contacts authorization is still `.notDetermined` in CLI usage.

## Root cause

`ContactResolver.create()` awaited `CNContactStore.requestAccess(for: .contacts)` whenever Contacts auth was undecided. In CLI/headless contexts that callback can fail to resolve promptly, so read commands that only need optional contact-name enrichment can appear to hang even after Messages DB work finished.

## Fix

- Add an explicit `ContactsAccessPolicy` with the existing default behavior preserved as `requestIfNeeded`.
- Use `skipIfNotDetermined` only for `chats` and `history`, so those read commands fail open without contact-name enrichment.
- Keep `send`, RPC, bridge intro, and other explicit contact/name-resolution paths on the default requesting resolver.
- Add resolver tests for both branches: skip policy does not request Contacts, default policy still does.
- Add an `0.11.1` changelog entry thanking @cemendes.

Closes #135.

## Verification

- `swift test --filter ContactResolver` passed: 4 tests.
- `swift test --filter 'ChatHistory|ContactResolver|sendCommandResolvesUniqueContactName|sendCommandRejectsAmbiguousContactName|accountNickname'` passed: 26 tests, including send/name-resolution coverage.
- `make build` passed and produced `bin/imsg` plus `bin/imsg-bridge-helper.dylib`; only existing helper deprecation warnings were emitted.
- Contacts status on this Mac was `.notDetermined` (`CNContactStore.authorizationStatus(for: .contacts).rawValue == 0`).
- Live proof under `.notDetermined`: `bin/imsg chats --limit 1 --json` returned within a 10s alarm; sanitized output showed one chat and no `contact_name` field.
- Live proof under `.notDetermined`: `bin/imsg history --chat-id <id> --limit 1 --json` returned within a 10s alarm; sanitized output showed one message and no `sender_name` field.
- `make lint` exited 0; repo has existing warning-level file length/line length violations, no serious findings.
- `make test` passed: 323 tests.

## Maintainer note

Local autoreview was attempted but the review engine returned empty output after a long-running child process was stopped, so it is not counted as proof here. The patch was manually reviewed against all `ContactResolver.create` call sites: only `chats` and `history` opt into no-prompt behavior; send/RPC/name paths still use the default requesting behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 0.11.1 - Unreleased
 
+### Read Commands
+- fix: let `chats` and `history` run without prompting for Contacts when permission is still undecided, while preserving Contacts prompts for explicit name-resolution flows (#135, thanks @cemendes).
+
 ## 0.11.0 - 2026-05-31
 
 ### Send
--- a/Sources/IMsgCore/ContactResolver.swift
+++ b/Sources/IMsgCore/ContactResolver.swift
@@ -34,6 +34,11 @@ public final class NoOpContactResolver: ContactResolving, Sendable {
   public func searchByName(_ query: String) -> [ContactMatch] { [] }
 }
 
+public enum ContactsAccessPolicy: Sendable {
+  case requestIfNeeded
+  case skipIfNotDetermined
+}
+
 public final class ContactResolver: ContactResolving, @unchecked Sendable {
   #if os(macOS)
     private let phoneToName: [String: String]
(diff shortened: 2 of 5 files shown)


steipete:
@clawsweeper re-review

Pushed maintainer fixup at `a783922`.

What changed since the blocked review:
- The global `.notDetermined` fail-open was narrowed behind `ContactsAccessPolicy`.
- Only `chats` and `history` use `skipIfNotDetermined`.
- `send`, RPC, bridge intro, and other explicit contact/name-resolution flows still use default `requestIfNeeded` behavior.
- Added resolver regression tests for both policies.
- Added live proof on this Mac with Contacts status `.notDetermined` (`rawValue == 0`): both `bin/imsg chats --limit 1 --json` and `bin/imsg history --chat-id <id> --limit 1 --json` returned within a 10s alarm, with sanitized output showing no contact-name enrichment.

Local proof: `make build`, focused Swift tests, `make lint` (exit 0, existing warnings only), and `make test` (323 tests) passed.

PR #138, opened 2026-06-07
fix(bridge): defer dylib init off the constructor (macOS 26 dyld init-order)

omarshahine:
## Summary

Move all ObjC/Foundation/IMCore work out of the injected dylib's `__attribute__((constructor))` (`injectedInit`) into `bridgeBootstrap()`, dispatched on the main queue. The constructor now only enqueues (a libdispatch call, no synchronous ObjC message dispatch).

**Why:** macOS 26 tightened dyld initializer ordering for platform/system apps. Touching ObjC/Foundation/IMCore at constructor time can run before libSystem finishes bootstrapping ("dyld initialized but libSystem has not") and abort Messages.app on launch. Deferring onto the main queue runs the work only after the process is fully initialized.

Based on `v0.11.0`; constructor change only (+16/-2).

## Verification

- `make build-dylib` clean (arm64e).
- **Live regression test on macOS 26.4.1:** built natively, injected via `imsg launch --kill-only` + `--dylib <built>`. `vmmap` confirmed the new dylib mapped into the live Messages process; `.imsg-bridge-ready` rewritten; `imsg status --json` -> `advanced_features:true, v2_ready:true, "Connected to Messages.app"`; `imsg chats` RPC roundtrip works; selectors identical to the released dylib (plus a bonus `pollPayloadMessage:true`). No crash, bridge fully up. Rolled back to the released dylib afterward.

## Scope / caveat

This proves **no regression on macOS 26.4.x**, where injection already works. It does **not** prove it fixes the macOS 26.5 injection/crash path (no 26.5 host with a working bridge available to test) — that remains the motivating hypothesis. The change is sound hardening regardless and is risk-free on hosts where the dylib already loads.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 ### Local Lookups
 - feat: add `--local` modes for `account`, `whois`, and `nickname` so common introspection can read local history or Contacts without launching the IMCore bridge (#132, thanks @ranaroussi).
 
+### Advanced IMCore
+- fix: defer injected bridge bootstrap until after Messages startup so macOS 26 dyld constructor ordering cannot touch ObjC/Foundation/IMCore before the process is ready (#138, thanks @omarshahine).
+
 ## 0.10.0 - 2026-05-28
 
 ### Watch
--- a/Sources/IMsgHelper/IMsgInjected.m
+++ b/Sources/IMsgHelper/IMsgInjected.m
@@ -4375,56 +4375,74 @@ static void startV2InboxWatcher(void) {
 // tightened dyld initializer ordering for platform/system apps, so touching
 // ObjC/Foundation/IMCore at constructor time can execute before libSystem has
 // finished bootstrapping ("dyld initialized but libSystem has not") and abort
-// Messages.app on launch. injectedInit() defers this onto the main queue, which
-// only drains once the process is fully initialized.
+// Messages.app on launch. injectedInit() only schedules this delayed bootstrap;
+// the lock file is written after the watchers are installed below.
 static void bridgeBootstrap(void) {
-    NSLog(@"[imsg-bridge] Dylib injected into %@", [[NSProcessInfo processInfo] processName]);
+    static dispatch_once_t once;
+    dispatch_once(&once, ^{
+        @autoreleasepool {
+            initFilePaths();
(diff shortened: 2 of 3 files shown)


steipete:
Maintainer pass on PR #138 complete.

What changed beyond the original patch:
- Kept the dylib constructor ObjC/Foundation/IMCore-free. It now performs only a libdispatch main-queue enqueue.
- Starts the existing 2s startup delay from the first main-queue turn, not from dyld constructor time, so the bootstrap cannot become eligible before Messages first services the main queue.
- Runs bridge bootstrap exactly once under `dispatch_once` and an explicit `@autoreleasepool`.
- Leaves the ready marker behavior behind `startFileWatcher()`, after watcher setup, and keeps v2 watcher/event observer startup in the same bootstrap boundary.
- Added a source-shape regression test for the constructor boundary.

Proof run locally on macOS 26.5 build 25F71:
- `make lint` passed. Existing warnings only; no serious lint failures.
- `make test` passed: 322 tests.
- `make build` passed: built `bin/imsg` and universal `bin/imsg-bridge-helper.dylib`.
- `make build-dylib` passed: built `.build/release/imsg-bridge-helper.dylib`.
- `autoreview --mode local` first found the constructor-timer timing issue; after the fix, rerun was clean with no accepted/actionable findings.

Live injection caveat:
- Local host is macOS 26.5 build 25F71, but SIP is enabled.
- Checked approved alternate targets: Mac Studio 26.5 25F71, clawmac 26.5.1 25F80, and Parallels `macOS Tahoe` 26.5 25F71. All have SIP enabled.
- Because no SIP-disabled macOS 26.5 target was available, I could not produce a live 26.5 DYLD injection/RPC proof.
- On the local 26.5 host, `imsg status --json` reported `advanced_features:false` and `v2_ready:false`; `imsg launch --dylib .build/release/imsg-bridge-helper.dylib --json` refused with `sip_enabled` before injection; Messages stayed running with the same PID, no `.imsg-bridge-ready` marker existed, and `vmmap` showed no custom dylib mapping. A redacted `imsg chats --limit 1 --json` smoke was also blocked by Full Disk Access for this Codex parent process, so no private content was read.

CI on pushed head `9b3bdaebf1f2e3ee6f6de99245f74c87ced7bd8e` is green:
- `linux-read-core` passed.
- `macos` passed.
- Socket checks passed.

PR #142, opened 2026-06-10
fix: coalesce URL preview split sends

### LINKED ISSUE #141: Coalesce Apple URL-preview split-sends into a single notification
## Summary

Apple delivers a URL-preview "split-send" as **two** `chat.db` rows: the user's text row (e.g. `Dump https://example.com`) and a separate `URLBalloonProvider` preview row, arriving ~0.8–2.0s apart. Consumers that want this as one logical message (e.g. OpenClaw) currently have to **debounce and merge client-side**, which is imprecise (you can't structurally tell a split-send from two separate sends without the balloon marker) and adds latency to every DM.

Now that imsg exposes `balloon_bundle_id` (#137), imsg has the structural signal to coalesce these rows **at the source** and emit a single logical message — removing the need for any client-side debounce/merge.

## Proposal

When imsg surfaces messages (watch notifications and history/search reads), coalesce an Apple URL-preview balloon row into its originating text row so consumers receive **one** message instead of two:

- Detect the `com.apple.messages.URLBalloonProvider` balloon row.
- Associate it with the immediately preceding same-chat / same-sender text row it previews.
- Emit a single coalesced message (preserve the text row's GUID for reply threading; annotate/attach the preview as appropriate).
- Apply consistently across `watch`, `messages`, and `searchMessages` so all read paths agree.

## Why imsg, not the client

- imsg owns the chat.db read path and already has `balloon_bundle_id` plus GUID/threading context — it's the natural owner of "un-splitting" Apple's split-send.
- Removes per-consumer reimplementation. OpenClaw currently does this in a debouncer that holds **every** DM for 2.5s so the preview row can arrive (see openclaw/openclaw#90858).
- Eliminates the consumer-side latency tradeoff and the "did this build emit the field?" compatibility branching.
- Single source of truth: one logical message, structurally derived, not timing-guessed.

## Context / related

- #137 — exposes `balloon_bundle_id` (the metadata this builds on)
- openclaw/openclaw#90858 — interim client-side gate consuming the metadata
- openclaw/openclaw#91243 — tracks deleting the OpenClaw client-side coalesce path once this lands

## Acceptance criteria

- A `Dump https://example.com` split-send is emitted as one message across watch / history / search.
- Two genuinely separate same-sender sends are NOT coalesced.
- Reply threading (GUID) and timestamps remain correct on the coalesced message.
- Tests cover: URL split-send coalesced; non-URL balloons (polls, etc.) untouched; separate sends untouched.

steipete:
Closes #141.

## Summary
- Coalesce Apple `com.apple.messages.URLBalloonProvider` split-send rows into one logical message for history, search, and watch/messagesAfter.
- Preserve raw `balloon_bundle_id` while adding nested `url_preview` metadata for the folded preview row.
- Keep public/watch cursors moving across suppressed previews, preserve logical limits after coalescing, and leave standalone/duplicate URL balloons covered by the existing dedupe path.
- Split large model/query helpers while keeping behavior local to existing `MessageStore` boundaries.

## Proof
- `make format`
- `make lint` (passes; existing non-serious warnings only)
- `swift test --filter URLPreview`
- `make test` (349 tests)
- `make build` (builds `bin/imsg` and `bin/imsg-bridge-helper.dylib`; existing `NSKeyedUnarchiver unarchiveObjectWithData:` deprecation warning only)
- `/Users/steipete/Projects/agent-skills/skills/autoreview/scripts/autoreview --mode local` clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### JSON Output
 - feat: expose raw Messages `balloon_bundle_id` in message JSON/RPC payloads so consumers can recognize URL-preview rows without parsing text (#137, thanks @omarshahine).
+- fix: coalesce Apple URL-preview split-send rows into one logical message across history, search, and watch, with `url_preview` metadata when the preview row is folded into the text row (#141, thanks @omarshahine).
 
 ### Advanced IMCore
 - fix: defer injected bridge bootstrap until after Messages startup so macOS 26 dyld constructor ordering cannot touch ObjC/Foundation/IMCore before the process is ready (#138, thanks @omarshahine).
--- a/README.md
+++ b/README.md
@@ -159,6 +159,7 @@ Message objects include:
 - `reply_to_text`, `reply_to_sender` (parent body + handle for threaded
   replies and non-reaction associations, when the parent row is still in
   chat.db)
+- `balloon_bundle_id`, `url_preview`
 - `sender`, `sender_name`, `is_from_me`, `text`, `created_at`
 - `attachments`, `reactions`
 - `poll` for native Apple Messages poll creation and vote rows
--- a/Sources/IMsgCore/Message+URLPreview.swift
+++ b/Sources/IMsgCore/Message+URLPreview.swift
@@ -1,0 +1,49 @@
+import Foundation
+
+extension Message {
+  public struct URLPreviewMetadata: Sendable, Equatable {
+    public let rowID: Int64
+    public let guid: String
+    public let balloonBundleID: String
+    public let date: Date
+
+    public init(rowID: Int64, guid: String, balloonBundleID: String, date: Date) {
+      self.rowID = rowID
+      self.guid = guid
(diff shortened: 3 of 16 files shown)


steipete:
Merged as 1c5160947c92a06ebf12b2a8d7c28104945279eb.

Proof run before merge:
- `make format`
- `make lint` (passes; existing non-serious warnings only)
- `swift test --filter URLPreview` (19 URL-preview-focused tests)
- `make test` (349 tests)
- `make build` (built `bin/imsg` and `bin/imsg-bridge-helper.dylib`; existing `NSKeyedUnarchiver unarchiveObjectWithData:` deprecation warning only)
- `/Users/steipete/Projects/agent-skills/skills/autoreview/scripts/autoreview --mode local` clean

CI:
- `linux-read-core` passed in 1m13s
- `macos` passed in 2m3s

Caveat: no live Messages injection was needed for this chat.db/history/search/watch coalescing fix; validation is fixture-backed local tests plus CI.

PR #143, opened 2026-06-21
fix: render text formatting on the RPC `send` bridge path

omarshahine:
## Problem

`handleSend` (RPC method `send`) forwards `text` and `file` to the IMCore bridge but **never the attributed-text format ranges**. So direct/handle sends come through plain even on macOS 15+ — only `send-rich` (CLI) and the RPC `sendRich` handler honor formatting.

This bites the OpenClaw gateway: its `message` tool routes plain handle targets (`to: +1…`) through the `send` method and puts format ranges in a `formatting` param. imsg's `send` ignores it, so `**bold**`/`*italic*` notifications silently lose styling. Replies render correctly only because the gateway routes those through `sendRich`.

## Fix

Thread the format ranges through `handleSend` → `sendViaBridge` → the bridge `sendMessage` action, mirroring what `sendRich` already does. Accept `formatting` (the key OpenClaw's `message` tool emits) alongside `text_formatting`/`textFormatting` for parity with `handleSendRich`.

- Bridge transport only; AppleScript sends stay plain (attributed text needs the private API).
- No behavior change when no formatting is supplied.

## Tests

- `rpcSendThreadsTextFormattingToBridge` — asserts ranges from the `formatting` key reach `invokeBridge` as `textFormatting`.
- `rpcSendWithoutFormattingOmitsTextFormatting` — asserts no `textFormatting` key when none is supplied.

All `RPCServerBridge` tests pass (7/7, no regressions).

## End-to-end proof (real device, macOS 15+, IMCore bridge v2)

Same RPC `send` call — method `send`, bare `formatting` key, `transport: bridge`, same self-chat (`lobster@shahine.com`) — run through both binaries. Only the binary differs. Verified by reading back the `attributedBody` Messages persisted in `chat.db` and decoding the attribute runs.

**Request (identical except the binary that serves it):**
```json
{"method":"send","params":{"to":"lobster@shahine.com","transport":"bridge",
 "text":"… bold test …","formatting":[{"start":0,"length":N,"styles":["bold"]}]}}
```

**BEFORE — unfixed `/opt/homebrew/bin/imsg` (0.11.1):**
```
RPC response: {"ok":true,"transport":"bridge","guid":"546F64D8-…"}
chat.db attributedBody runs:
  range 0..<30 "BEFORE bold test (brew 0.11.1)"  style=(none)   ← formatting dropped, plain
```

**AFTER — this branch's release build (`bin/imsg`):**
```
RPC response: {"ok":true,"transport":"bridge","guid":"58A3EE57-…"}
chat.db attributedBody runs:
  range 0..<5  "AFTER"                    style=__kIMTextBoldAttributeName  ← bold applied
  range 5..<28 " bold test (branch fix)"  style=(none)
```

The unfixed `send` persists no bold attribute; the fixed `send` persists `__kIMTextBoldAttributeName` over exactly the requested range. `send-rich` already produced this; the fix brings the plain `send` path (the one OpenClaw's `message` tool uses for `to:` handle targets) to parity. (Lobster is a headless Mac, so this is decoded persisted styling rather than a UI screenshot.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 0.11.2 - Unreleased
 
+### Send
+- fix: thread attributed-text formatting through the RPC `send` bridge path, not just `send-rich`, so direct/handle sends render **bold**/*italic*/etc. on macOS 15+. `handleSend` now forwards format ranges to the bridge, accepting `formatting` (the key OpenClaw's `message` tool emits) alongside `text_formatting`/`textFormatting`.
+
 ## 0.11.1 - 2026-06-10
 
 ### Read Commands
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## 0.11.2 - Unreleased
 
 ### Send
-- fix: thread attributed-text formatting through the RPC `send` bridge path, not just `send-rich`, so direct/handle sends render **bold**/*italic*/etc. on macOS 15+. `handleSend` now forwards format ranges to the bridge, accepting `formatting` (the key OpenClaw's `message` tool emits) alongside `text_formatting`/`textFormatting`.
+- fix: thread attributed-text formatting through the RPC `send` bridge path, not just `send-rich`, so direct/handle sends render **bold**/*italic*/etc. on macOS 15+. `handleSend` now forwards format ranges to the bridge, accepting `formatting` (the key OpenClaw's `message` tool emits) alongside `text_formatting`/`textFormatting` (#143, thanks @omarshahine).
 
 ## 0.11.1 - 2026-06-10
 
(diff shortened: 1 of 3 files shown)


omarshahine:
@clawsweeper re-review with updated proof

steipete:
Maintainer validation complete on macOS with Swift 6.4:

- `git diff --check`
- `make lint` (passes; existing non-serious warnings only)
- `make test` (all Swift Testing suites pass, including the two new RPC formatting regressions)
- GitHub CI: `macos` and `linux-read-core` both green

The contributor supplied live before/after bridge proof in the PR body. I am explicitly accepting the focused unreleased changelog entry here because the fix is user-visible and the maintainer workflow requires it; it also preserves contributor credit.

PR #145, opened 2026-06-23
fix: thread rpc send replies

### LINKED ISSUE #144: RPC send method ignores reply_to parameter — sets reply_to_guid to wrong message, never sets thread_originator_guid
## Description

The JSON-RPC `send` method accepts a `reply_to` parameter but does not use it correctly. On imsg 0.11.1, the RPC `send` method:

1. **Ignores the specified `reply_to` GUID** — instead of threading to the specified message, it sets `reply_to_guid` to the most recent message in the chat (regardless of which GUID was passed).
2. **Never sets `thread_originator_guid`** — the `thread_originator_guid` field is always empty for RPC `send` messages.

In contrast, the `imsg send-rich --reply-to <guid>` CLI command correctly sets both `thread_originator_guid` and `reply_to_guid` to the specified message.

## Steps to Reproduce

1. Start `imsg rpc --json`
2. Send a message with an explicit `reply_to` pointing to an older message (not the most recent):
```json
{"jsonrpc":"2.0","id":1,"method":"send","params":{"text":"test","to":"+15551234567","reply_to":"OLDER-MESSAGE-GUID-HERE"}}
```
3. Check the sent message in `chat.db` or via `imsg history`

## Expected Behavior

The sent message should have:
- `thread_originator_guid` = the GUID specified in `reply_to`
- `reply_to_guid` = the GUID specified in `reply_to`

(Same behavior as `imsg send-rich --reply-to <guid>`)

## Actual Behavior

The sent message has:
- `thread_originator_guid` = (none)
- `reply_to_guid` = the most recent message GUID in the chat (not the one specified in `reply_to`)

## Environment

- imsg 0.11.1
- macOS 15.5 (Tahoe)
- SIP disabled, IMCore bridge connected
- Apple Silicon (aarch64)

## Workaround

Use `imsg send-rich --chat <guid> --reply-to <guid> --text <text>` via the CLI instead of the RPC `send` method. The `send-rich` command correctly threads replies via the IMCore bridge.

## Context

- Issue #65 requested `--reply-to` for `imsg send` (closed as not planned for AppleScript path).
- Issue #108 fixed the RPC `send` transport to use the IMCore bridge, but did not address `reply_to` threading.
- The `send-rich` CLI command already supports `--reply-to` correctly via the IMCore bridge.
- The RPC `send` method already uses the bridge (since v0.8.1), so it should be able to pass `reply_to` through to the bridge send path, similar to how `send-rich` does it.

TurboTheTurtle:
Closes #144.

Plain RPC `send` now forwards reply target aliases into the bridge `selectedMessageGuid` payload, matching `send.rich` behavior.

Reply-target sends now fail instead of falling back to unthreaded AppleScript delivery when the bridge is unavailable or fails.

Proof note: on a macOS arm64 branch-built `imsg` with SIP enabled and no bridge helper (`bridge_version: 0`), RPC `send` with `reply_to` returns `Invalid params: reply_to requires bridge transport; AppleScript fallback cannot send threaded replies`, so the no-bridge path fails before AppleScript delivery. Positive threaded-send proof has been requested from Mantis because this local machine cannot inject the bridge.

Verified with focused RPC bridge coverage plus the full test and lint gates.

### DIFF
--- a/Sources/IMsgHelper/IMsgInjected.m
+++ b/Sources/IMsgHelper/IMsgInjected.m
@@ -825,6 +825,7 @@ static NSDictionary* handleStatus(NSInteger requestId, NSDictionary *params) {
         @"read_available": @(hasRegistry),
         @"bridge_version": @2,
         @"v2_ready": @(rpcInboxTimer != nil),
+        @"attachment_metadata": @YES,
         @"selectors": selectors
     });
 }
--- a/Sources/imsg/RPCServer+Handlers.swift
+++ b/Sources/imsg/RPCServer+Handlers.swift
@@ -173,6 +173,9 @@ extension RPCServer {
     }
     let transport = try RPCSendTransport.parse(stringParam(params["transport"]))
     let region = stringParam(params["region"]) ?? "US"
+    let selectedMessageGuid = stringParam(
+      params["reply_to"] ?? params["replyTo"] ?? params["reply_to_guid"] ?? params["message_guid"]
+    ).flatMap { $0.isEmpty ? nil : $0 }
     let rawRecipient = stringParam(params["to"]) ?? ""
     let rawInput = ChatTargetInput(
       recipient: rawRecipient,
@@ -262,7 +265,8 @@ extension RPCServer {
         let data = try await sendViaBridge(
           chatGUID: bridgeChatGUID,
           text: text,
-          file: file
+          file: file,
+          selectedMessageGuid: selectedMessageGuid
         )
         var result: [String: Any] = ["ok": true, "transport": "bridge"]
         if let guid = data["messageGuid"] as? String, !guid.isEmpty {
@@ -278,16 +282,20 @@ extension RPCServer {
         respond(id: id, result: result)
         return
       } catch let err as RPCError {
-        if transport == .bridge {
+        if transport == .bridge || selectedMessageGuid != nil {
           throw err
         }
       } catch {
-        if transport == .bridge {
+        if transport == .bridge || selectedMessageGuid != nil {
           throw RPCError.internalError(String(describing: error))
         }
       }
     } else if transport == .bridge {
       throw RPCError.invalidParams("bridge transport requires an existing chat target")
+    } else if selectedMessageGuid != nil {
+      throw RPCError.invalidParams(
+        "reply_to requires bridge transport; AppleScript fallback cannot send threaded replies"
+      )
     }
 
     try sendMessage(options)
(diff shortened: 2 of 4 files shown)


TurboTheTurtle:
@clawsweeper re-review

TurboTheTurtle:
Real behavior proof update for the fail-closed path.

Real environment tested: macOS arm64 local dev host, branch-built `imsg` 0.11.1 from `aef9e345d5e3`, SIP enabled, no bridge helper available (`bridge_version: 0`).

Exact steps or command run after this patch: ran JSON-RPC `send` over stdin with `reply_to` set and no injected bridge.

Evidence after fix:
```json
{"error":{"code":-32602,"data":"reply_to requires bridge transport; AppleScript fallback cannot send threaded replies","message":"Invalid params"},"id":"proof-no-bridge","jsonrpc":"2.0"}
```

Observed result after fix: the request failed before AppleScript fallback, so the branch did not send an unthreaded fallback message when it could not honor `reply_to`.

Positive threaded-send proof still needs a SIP-disabled/injected Messages bridge environment; this host reports SIP enabled and `bridge_version: 0`.

TurboTheTurtle:
@openclaw-mantis visual task: verify imsg rpc send with reply_to threads to the selected message and bridge-unavailable reply sends error without AppleScript fallback.

TurboTheTurtle:
@clawsweeper re-review

steipete:
Maintainer closeout:

- Forwarded `reply_to`, `replyTo`, `reply_to_guid`, and `message_guid` through the standard RPC `send` path.
- Kept threaded sends fail-closed when the IMCore bridge is unavailable, so they cannot silently become unthreaded AppleScript sends.
- Added caption + formatting support for threaded attachments and an explicit helper capability probe, preventing older injected helpers from silently dropping caption/reply metadata.
- Preserved contributor credit in the maintainer fixup commit.

Proof:

- `swift test --filter RPCServerBridgeTests` — 12 passed.
- `make test` — 361 passed.
- `make lint` — passed (existing warnings only).
- `swift build --product imsg` — passed.
- Fresh Codex autoreview — clean after both accepted findings were fixed.
- Live RPC attempt against the maintainer-provided test chat: threaded send failed closed with the expected bridge-required error and did not fall back. A real AppleScript baseline send was attempted but macOS denied Automation access with `-1743`; no message was sent.

Thanks @TurboTheTurtle for finding and fixing the missing standard-send reply plumbing.

PR #149, opened 2026-07-01
fix(bridge): restore chat deletion on macOS 26

### LINKED ISSUE #146: chat-delete fails with `Dylib error: deleteChat: not available` on macOS 26.5.1
## Summary

`imsg chat-delete` is exposed by the CLI and `chats.delete` is exposed by RPC, but both fail at runtime on my machine with:

```text
Dylib error: deleteChat: not available
```

Other bridge-backed chat mutations work on the same conversation, so this looks specific to the `deleteChat` selector path rather than a general injection failure.

## Environment

- `imsg 0.11.1`
- macOS `26.5.1` (`25F80`)
- SIP: disabled
- Installed via Homebrew (`steipete/tap/imsg`)

## Reproduction

1. Launch Messages with injection:

```bash
imsg launch
```

2. Confirm bridge status:

```bash
imsg status --json
```

Relevant output:

```json
{
  "version":"0.11.1",
  "advanced_features":true,
  "bridge_version":2,
  "v2_ready":true,
  "typing_indicators":true,
  "read_receipts":true,
  "rpc_methods":["chats.delete","chats.markUnread", "message.delete", ...]
}
```

3. Attempt to delete an existing chat:

```bash
imsg chat-delete --chat 'any;-;+14793086064' -v
```

Observed result:

```text
error: Dylib error: deleteChat: not available
```

4. Same failure via RPC:

```bash
printf '%s\n' '{"jsonrpc":"2.0","id":"1","method":"chats.delete","params":{"chat_guid":"any;-;+14793086064"}}' | imsg rpc
```

Observed result:

```json
{"id":"1","error":{"code":-32603,"message":"Internal error","data":"Dylib error: deleteChat: not available"},"jsonrpc":"2.0"}
```

## Additional observations

On the same SMS thread (`guid: any;-;+14793086064`), `chat-mark` works:

```bash
imsg chat-mark --chat 'any;-;+14793086064' --unread
imsg chat-mark --chat 'any;-;+14793086064' --read
```

Both succeeded.

So the bridge is connected and chat mutation is functional in general, but `deleteChat` specifically is unavailable.

## Likely cause

From the source, `chat-delete` appears to call `IMChatRegistry.deleteChat:` in the injected helper. On this macOS build, that selector appears not to be present at runtime, triggering:

```text
deleteChat: not available
```

Is `chat-delete` expected to work on macOS 26.x, or does it need a different selector / fallback path on newer systems?

steipete:
## What Problem This Solves

`chat-delete` fails on macOS 26 because `IMChatRegistry` no longer exposes the older `deleteChat:` selector. This fixes #146.

## Why This Change Was Made

The injected helper now prefers the existing `deleteChat:` path when present and falls back to the macOS 26 `_chat_remove:` registry path. It fails closed when neither selector exists, advertises both capabilities in bridge status, and reports the method used for diagnostics.

## User Impact

Users on macOS 26 can delete chats through the existing bridge command again. Older macOS behavior remains unchanged.

## Evidence

- Live Objective-C runtime probe on macOS 26.5.2: `deleteChat:` absent; `_chat_remove:` present on `IMChatRegistry` with the expected one-object argument signature.
- [BlueBubbles IMCore documentation](https://docs.bluebubbles.app/private-api/imcore-documentation) independently identifies `_chat_remove:` as the registry chat-deletion path.
- `make lint` — passed (existing warnings only).
- `make test` — 363 tests passed.
- `make build-dylib` — passed (one existing deprecation warning).
- Fresh autoreview — no actionable findings, 0.94 confidence.

A destructive live delete was intentionally not run because this host has no disposable conversation and SIP is enabled, so the injected bridge cannot be launched here. The runtime selector proof is live and non-destructive.

### DIFF
--- a/Sources/IMsgHelper/IMsgInjected.m
+++ b/Sources/IMsgHelper/IMsgInjected.m
@@ -814,7 +814,11 @@ static NSDictionary* handleStatus(NSInteger requestId, NSDictionary *params) {
         @"retractMessagePart": @(gHasRetractMessagePart),
         @"sendMessageReason": @(gHasSendMessageReason),
         @"pollPayloadMessage": @(pollPayloadMessageInitializerAvailable()),
-        @"pollVoteMessage": @(pollVoteMessageInitializerAvailable())
+        @"pollVoteMessage": @(pollVoteMessageInitializerAvailable()),
+        @"deleteChat": @(hasRegistry &&
+            [registryClass instancesRespondToSelector:NSSelectorFromString(@"deleteChat:")]),
+        @"removeChat": @(hasRegistry &&
+            [registryClass instancesRespondToSelector:NSSelectorFromString(@"_chat_remove:")])
     };
 
     return successResponse(requestId, @{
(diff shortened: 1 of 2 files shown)


steipete:
Land-ready at exact head `25c09c1799cac72bfaefa2d7f525a0bce66cebf2`.

- Live macOS 26.5.2 runtime probe: `deleteChat:` absent; `_chat_remove:` present with the expected one-object signature.
- Focused regression: passed.
- `make lint`: passed (existing warnings only).
- `make test`: 363 tests passed.
- `make build-dylib`: passed (one existing deprecation warning).
- Fresh autoreview: no actionable findings (0.94 confidence).
- GitHub Actions run 28540334742: macOS and Linux lanes passed; Socket checks passed.

Known proof gap: no destructive chat deletion was run because this host has no disposable conversation and SIP prevents injected-bridge launch. The selector contract itself was verified live and non-destructively.

PR #153, opened 2026-07-03
fix: resolve poll update option text

### LINKED ISSUE #152: Resolve poll vote option text from native option update rows
Native Apple Messages polls can have choices added after the poll is created. Those choices are stored in poll update rows, not necessarily in the original poll creation row.

Today `imsg` can expose `poll.vote.option_id` for those votes, but `poll.vote.option_text` may be missing when the selected option was added later and the vote row still points back to the original poll GUID.

Expected behavior:

- History and watch output should resolve `poll.vote.option_text` for original poll choices.
- They should also resolve `poll.vote.option_text` for choices added later through native Add Choice updates.
- Unknown or malformed update payloads should not break poll readback.
- Existing poll JSON fields should remain compatible.

This came out of follow-up review on native poll support after testing add-choice and vote flows against real Messages rows.

veteranbv:
Fixes #152.

Native poll choices added after creation are stored in poll update rows. A vote row can still point back to the original poll GUID, so resolving option text from only the original creation row can leave `poll.vote.option_text` empty even though `poll.vote.option_id` is present.

This updates poll option resolution to read the original poll row plus native option update rows associated to that poll, then merge option IDs in database order. If the schema does not expose associated-message columns, it keeps the existing exact-GUID lookup path.

Tests add the missing shape: original poll row, later native Add Choice update row, and a vote row associated to the original poll GUID selecting the added option.

Verification:

- `swift format lint --recursive Sources Tests TestsLinux`
- `swift test --filter MessagePollTests`
- `swift test`
- `make test`

Local note: `make lint` could not complete because this machine does not have `swiftlint` installed (`make: swiftlint: No such file or directory`). The `swift format` portion passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.12.2 - Unreleased
 
 ### Native Polls
+- fix: resolve native poll vote option text from Add Choice update rows when votes still reference the original poll message (#152, #153, thanks @veteranbv).
 - fix: restore native poll vote delivery on macOS 26.4 by persisting the Polls balloon and payload across the responding message objects (#150, thanks @omarshahine).
 
 ## 0.12.1 - 2026-07-02
--- a/Sources/IMsgCore/MessagePolls.swift
+++ b/Sources/IMsgCore/MessagePolls.swift
@@ -309,6 +309,24 @@ extension MessagePollEvent {
       metadata: metadata
     )
   }
+
+  func resolvingPollReference(pollGUID: String?, originalGUID: String?) -> MessagePollEvent {
+    if self.pollGUID == pollGUID, self.originalGUID == originalGUID {
+      return self
+    }
+    return MessagePollEvent(
+      kind: kind,
+      pollGUID: pollGUID,
+      question: question,
+      options: options,
+      vote: vote,
+      votes: votes,
+      originalGUID: originalGUID,
+      creator: creator,
+      participants: participants,
+      metadata: metadata
+    )
+  }
 }
 
 private struct PollFacts {
--- a/Sources/IMsgCore/MessageStore+Polls.swift
+++ b/Sources/IMsgCore/MessageStore+Polls.swift
@@ -24,8 +24,9 @@ extension MessageStore {
   }
 
   /// Ordered options of the poll identified by `guid`, decoded from its
-  /// creation message. Used by `poll vote` to resolve a 1-based option index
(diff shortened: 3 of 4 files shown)


veteranbv:
Follow-up after retesting this against a real Messages database.

The first proof pass exposed one more Apple row shape I did not want to leave awkward for consumers. When a participant adds choices and votes, Messages can produce:

- the original poll creation row
- a native Polls update row with `associated_message_type=2`
- vote rows where one points at the original poll and another points at the update row

The earlier patch resolved `vote.option_text`, but a vote pointed at the update row could still surface `poll_guid` / `original_guid` as the update row instead of the source poll. Commit `9f3820d` fixes that. Votes that physically point at a poll update row now resolve back to the source poll for `poll_guid` and `original_guid`, while still using the updated option set to fill `vote.option_text`.

Redacted live proof from a fresh checkout at `9f3820d`:

```json
{
  "test_poll_found": true,
  "source_created_poll": {
    "event": "imessage.poll.created",
    "options_count": 2,
    "poll_guid_present": true
  },
  "update_rows": {
    "count": 1,
    "option_counts": [4]
  },
  "related_vote_rows": {
    "count": 2,
    "vote_to_original_count": 2,
    "vote_to_update_count": 0,
    "with_option_text": 2,
    "examples": [
      {
        "relation": "vote_to_original",
        "event": "imessage.poll.voted",
        "original_guid_present": true,
        "poll_guid_present": true,
        "poll_guid_matches_source_poll": true,
        "original_guid_matches_source_poll": true,
        "original_guid_matches_update_row": false,
        "vote": {
          "option_id_present": true,
          "option_text_present": true,
          "option_text_length": 10,
          "participant_present": true,
          "event_type": "selected"
        }
      },
      {
        "relation": "vote_to_original",
        "event": "imessage.poll.voted",
        "original_guid_present": true,
        "poll_guid_present": true,
        "poll_guid_matches_source_poll": true,
        "original_guid_matches_source_poll": true,
        "original_guid_matches_update_row": false,
        "vote": {
          "option_id_present": true,
          "option_text_present": true,
          "option_text_length": 10,
          "participant_present": true,
          "event_type": "selected"
        }
      }
    ]
  }
}
```

Validation run locally:

```text
swift format lint --recursive Sources Tests TestsLinux
swift test
make test
```

All three passed. `make lint` still cannot complete in my local environment because `swiftlint` is not installed; its `swift format` step passes before `swiftlint` is invoked.

@clawsweeper re-review

PR #154, opened 2026-07-04
docs(skill): expand imsg agent skill with full CLI surface

omarshahine:
## Summary

Rewrites `.agents/skills/imsg/SKILL.md` so agents get an accurate, complete picture of the CLI. The previous version covered only chats/history/search/send, hardcoded a machine-specific repo path (`~/Projects/imsg`), and never mentioned the standard-vs-bridge capability split.

Changes (verified against `imsg completions llm` and `docs/`):

- **Two-tier capability model up front**: standard commands (chats/group/history/watch/search/send/react + `--local` lookups) vs bridge commands (send-rich, tapback, poll, edit/unsend, chat-*), gated on `imsg status --json`, with explicit guidance to fall back to standard equivalents rather than suggest disabling SIP.
- **Progressive disclosure**: points agents at `imsg completions llm` for the full flag reference so the skill stays short.
- Keeps the original's key insight (resolve UI-visible people via `chats`/`contact_name`, not `search`) and adds what was missing: `group` before automating, `--start`/`--end` inclusive/exclusive semantics, `--since-rowid` as the watch cursor, `--convert-attachments`, SMS fallback behavior, and `react`'s most-recent-message-only limitation with `tapback` as the bridge alternative.
- Requires per-action user confirmation for destructive bridge commands (`unsend`, `delete-message`, `chat-delete`, `chat-leave`, `chat-remove-member`).
- Verification now uses `./bin/imsg` (the source-build output) for live read proof.

Review fix (540c803): default `nickname` moved to the bridge tier; the SIP-free tier and example now use `nickname --local`, with a note that `--local` returns your own AddressBook label rather than the iMessage-shared nickname. Matches `Sources/imsg/Commands/BridgeIntroCommands.swift` and the CLI help below.

**Poll docs scope (decoupled from #155):** an earlier revision documented the poll question echoing as a visible caption, the `--comment` override, and inbound question backfill. That behavior lives in #155, not on shipped `main` — `imsg poll --help` on 0.12.2 exposes only `--question`/`--option`, no `--comment`. Those claims were removed so this skill documents only shipped behavior and merges independently; #157 tracks re-adding them once #155 lands.

## Validation transcript

Run against imsg 0.12.2 (Homebrew) on macOS:

```console
$ imsg --version
0.12.2

$ imsg completions llm | grep -c "^### "   # command sections in the agent reference
33

$ imsg poll --help | grep -E '^\s+--'      # shipped poll surface: no --comment
  --db <value>	Path to chat.db (defaults to ~/Library/Messages/chat.db)
  --chat <value>	chat guid or rowid
  --chat-id <value>	chat rowid
  --question <value>	poll question
  --option <value>	poll option text; pass at least twice
  --option-id <value>	vote: UUID of the option to select
  --option-index <value>	vote: 1-based option number to select

$ imsg nickname --help | head -11
imsg nickname
Show contact-card / nickname info for a handle

The default mode reads the contact-card nickname the correspondent shared
over iMessage via the IMCore bridge (requires `imsg launch`, SIP disabled).

--local is a SIP-free alternative that returns YOUR local AddressBook
contact name for the handle. NOTE: this is a different datum — it is your
own contact label, not the iMessage-shared nickname/photo, which is only
available through the bridge.

$ imsg status --json | jq '{sip, v2_ready, basic_features, typing_indicators, read_receipts}'
{
  "sip": "enabled",
  "v2_ready": false,
  "basic_features": true,
  "typing_indicators": false,
  "read_receipts": false
}
```

(The `status` output above is from a SIP-enabled machine — exactly the state where the skill's fallback guidance applies.)

jq chat-filter example validated against synthetic NDJSON matching `docs/json.md`'s Chat schema:

```console
$ printf '%s\n' '{"id":1,"contact_name":"Beatrix Smith","identifier":"+15551234567"}' '{"id":2,"name":"Other"}' \
    | jq -s '.[] | select((.contact_name // .display_name // .name // .identifier // "" | ascii_downcase) | contains("beatrix")) | .id'
1
```

### DIFF
--- a/.agents/skills/imsg/SKILL.md
+++ b/.agents/skills/imsg/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: imsg
-description: "iMessage/SMS via the imsg CLI: read local Messages history, search, stream live, resolve contacts, and send (sends only when explicitly requested). Use for any Messages.app / chat.db task."
+description: "iMessage/SMS: local history, contacts, live watch, and requested sends."
 ---
 
 # imsg
@@ -72,10 +72,12 @@ Only after `imsg status` confirms the bridge is loaded (`imsg launch` injects it
 
 ```bash
 imsg send-rich --chat 'iMessage;-;+15551234567' --text 'hi' --reply-to MSG_GUID   # replies, effects, subjects
-imsg poll send --chat GUID --question 'Dinner?' --option 'Pizza' --option 'Sushi'  # native Messages poll; --question + 2+ --option
+imsg poll send --chat GUID --question 'Dinner?' --option 'Pizza' --option 'Sushi' --comment 'Vote by 5pm'
 imsg edit --chat GUID --message MSG_GUID --new-text 'updated'                     # macOS 13+
 imsg chat-create --addresses '+15551234567,+15559876543' --name 'Crew'
 ```
+
+`poll send` echoes `--question` as a best-effort plain caption after the Polls balloon; `--comment` overrides that caption. Do not retry automatically when only the caption fails: the poll may already be delivered. `history` and `watch` backfill a title-less inbound native poll's `poll.question` from its clean caption row.
 
 Destructive bridge commands — `unsend`, `delete-message`, `chat-delete`, `chat-leave`, `chat-remove-member` — need per-action user confirmation.
 
--- a/.agents/skills/imsg/SKILL.md
+++ b/.agents/skills/imsg/SKILL.md
@@ -77,7 +77,7 @@ imsg edit --chat GUID --message MSG_GUID --new-text 'updated'                   
 imsg chat-create --addresses '+15551234567,+15559876543' --name 'Crew'
 ```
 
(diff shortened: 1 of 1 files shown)


omarshahine:
@clawsweeper re-review

omarshahine:
Decoupled from #155 per maintainer decision. The poll caption echo, `--comment` override, and inbound question backfill live in #155, not shipped `main` (`imsg poll --help` on 0.12.2 exposes only `--question`/`--option`, no `--comment`). Those claims are removed (4e5aa72) so this skill documents only shipped behavior; #157 tracks re-adding them after #155 lands. PR body updated with the shipped-poll-surface proof.

@clawsweeper re-review

steipete:
Landed as e771b0ca88651e55975ce966470a94cbaa17f13c. Thank you, @omarshahine.

Maintainer proof:
- rebased onto current `main`
- YAML frontmatter parsed and bundled skill validator passed
- live `imsg poll --help` and `imsg completions llm` matched the documented caption and `--comment` behavior
- branch-wide autoreview found one unsafe fallback rule; fixed it, then reran to a clean result
- exact-head macOS and Linux CI passed

PR #155, opened 2026-07-05
fix(poll): make the poll question visible both directions (echo out, backfill in)

omarshahine:
## What Problem This Solves

Native iMessage poll balloons have **no title** — `item.title` is empty and Messages never renders it. The question you see on a native poll lives in a **separate caption message** whose `reply_to` is the poll guid (the composer's "comment or Send" field). imsg ignored this on both sides:

- **Outbound:** `poll send --question 'X'` stored `X` in the invisible `item.title`, so agent-created polls rendered as bare options with no question.
- **Inbound:** a native poll decoded from `chat.db` had `question=nil`, so agents saw a title-less `Poll — options: …`.

## Why This Change Was Made

Make the poll question visible in both directions, matching native behavior, with no consumer changes required (OpenClaw already renders `📊 Poll: <question>` when the field is present, and passes `--question` on send).

## User Impact

- **Outbound** — `poll send` / `poll.send` RPC now send `--question` as a **plain caption message** right after the poll — `associated_message_type = 0`, `thread_originator_guid = None`, i.e. **not a threaded reply**, so it renders as a normal caption bubble instead of the "reply connector" a threaded reply would draw. `--comment` overrides the echoed text; the echo is best-effort so a caption failure never fails an already-delivered poll. Agents keep passing only `--question`.
- **Inbound** — `enrichedPollEvent` **backfills** an empty created-poll question from its caption (earliest reply-to-the-poll message), so inbound native polls surface their question. imsg-created polls with a real title are unchanged; vote enrichment untouched.

## Evidence

Verified end-to-end on macOS 26 through a **live v2 bridge**, sending with this branch's build (`swift build --product imsg`).

**Outbound — live send** to a test group:
```
$ imsg poll send --chat-id 22 --question "PR#155 proof: which crustacean?" \
      --option Lobster --option Crab --option Shrimp --json
{"poll":{"kind":"created","question":"PR#155 proof: which crustacean?",
 "options":[Lobster,Crab,Shrimp]},"event":"imessage.poll.created",
 "messageGuid":"EAE66DE7-…","balloonBundleID":"…Polls"}
```

**Visual** — confirmed in the Messages UI on the receiving device: the poll shows the three options (Lobster / Crab / Shrimp), and directly beneath it the question appears as an ordinary caption bubble ("PR#155 proof: which crustacean?") — no threaded-reply connector line. (Screenshot available; can be attached to this PR.)

**Structural proof (chat.db)** — my sent caption is byte-for-byte the same shape as a **natively-composed** poll's caption, and differs from the earlier threaded-reply attempt that drew the connector:

```
poll_guid  origin          caption   assoc_type  thread_originator
EAE66DE7   this PR (sent)   01E1505B  0           NULL   ← clean, matches native
2E97B312   native (recv)    CCBD34D4  0           NULL   ← native reference
AC8A8E5B   native (recv)    1ECFCA87  0           NULL   ← native reference
A3DDC24F   old threaded try C615FCEF  100         A3DDC24F ← the rejected reply connector
```

Note: this bridge auto-populates `reply_to_guid` on every send (the poll itself got a `reply_to` although no `--reply-to` was passed), so both native and PR captions carry `reply_to`; the connector is driven by `thread_originator`/`assoc_type=100`, which the PR caption does **not** set — matching native exactly.

**Inbound — backfill** via `imsg history` (chat.db read, no bridge), native `[recv]` polls that decoded as `question=None` now surface their question:
```
[recv] [poll created] Favorite crustacean friend of a Lobster?: Lobster 🦞 / Crab 🦀 / Shrimp 🦐
[recv] [poll created] What color pill?: Red / Blue
[recv] [poll created] Pick a color: Black / White
```

**Review follow-ups addressed** (ClawSweeper P2s): CLI caption echo is now best-effort (`do/catch`, matching the RPC path); `pollCommentText` is guarded on `schema.hasReplyToGUIDColumn` so schemas without the column degrade to nil instead of throwing; poll-send tests record the full bridge call sequence and assert both `.sendPoll` and the `.sendMessage` caption echo. macOS + Linux CI green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 0.12.3 - Unreleased
 
+### Native Polls
+- fix: show native poll questions by sending a plain caption after created polls and backfilling empty inbound questions from clean native caption rows (#155, thanks @omarshahine).
+
 ### Packaging
 - fix: include and validate an arm64e slice in the injected bridge helper for macOS 26 Messages compatibility (#156, thanks @omarshahine).
 
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ Advanced IMCore (require `imsg launch` with SIP off — see
 - `imsg send-rich [--reply-to <guid>] [--file <path>]`,
   `imsg send-multipart`, `imsg send-attachment [--reply-to <guid>]`,
   `imsg tapback`
-- `imsg poll send (--chat <guid> | --chat-id <id>) --question <text> --option <text> --option <text> [--reply-to <guid>]`
+- `imsg poll send (--chat <guid> | --chat-id <id>) --question <text> [--comment <text>] --option <text> --option <text> [--reply-to <guid>]`
 - `imsg poll vote (--chat <guid> | --chat-id <id>) --poll <guid> (--option-id <id> | --option-index <n> | --option <text>)`
 - `imsg edit`, `imsg unsend`, `imsg delete-message`, `imsg notify-anyways`
 - `imsg chat-create`, `imsg chat-name`, `imsg chat-photo`,
@@ -139,6 +139,10 @@ Advanced IMCore (require `imsg launch` with SIP off — see
 `imsg status --json` reports native bridge selector capabilities. Poll creation
 requires `selectors.pollPayloadMessage`; poll voting requires
 `selectors.pollVoteMessage` plus `poll.vote` in `rpc_methods`.
+Messages does not render the poll payload title on the balloon, so `poll send`
+also sends a best-effort plain caption message right after the poll. The
+caption defaults to `--question`; use `--comment` when the visible text should
+be different from the stored question.
 
 `react` intentionally sends only the standard tapbacks Messages.app exposes
 reliably through automation. Custom emoji tapbacks can be read from
--- a/Sources/IMsgCore/MessagePolls.swift
+++ b/Sources/IMsgCore/MessagePolls.swift
@@ -327,6 +327,24 @@ extension MessagePollEvent {
       metadata: metadata
     )
   }
+
+  /// Returns a copy with `question` filled in. Used to backfill a native poll's
+  /// empty title (item.title) from its caption message so the poll is
+  /// self-describing to consumers.
+  func withQuestion(_ newQuestion: String) -> MessagePollEvent {
+    MessagePollEvent(
+      kind: kind,
+      pollGUID: pollGUID,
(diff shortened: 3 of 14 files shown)


omarshahine:
@clawsweeper re-review

Addressed all three P2 findings and added real behavior proof:
- CLI caption echo is now best-effort (do/catch, matching the RPC path) — a caption failure no longer exits nonzero after the poll delivered (36e1ece).
- `pollCommentText` is guarded on `schema.hasReplyToGUIDColumn`; a table without the column degrades to nil instead of throwing (8b8f451).
- Poll-send tests now record the full bridge call sequence and assert both `.sendPoll` and the `.sendMessage` caption echo (8b8f451).

Proof (PR body updated): live `imsg poll send` through a v2 bridge on macOS 26 using this branch's build; chat.db shows the sent caption is structurally identical to a natively-composed poll's caption (`assoc_type=0`, `thread_originator=None`) and differs from the old threaded-reply attempt (`assoc_type=100`, thread_originator set). Inbound native `[recv]` polls now surface their backfilled questions via `imsg history`. macOS + Linux CI green.

omarshahine:
Fixed the P1 (64ab396): `pollCommentText` now filters to clean native captions (`associated_message_type = 0 AND thread_originator_guid IS NULL`, schema-aware since both columns are optional), so a threaded inline reply to the poll can no longer be mistaken for the question. Added regression coverage: `messageStorePollBackfillsQuestionFromCleanCaption` (clean caption still backfills) and `messageStorePollBackfillIgnoresThreadedReply` (threaded reply ignored, question stays empty). macOS + Linux green.

@clawsweeper re-review

steipete:
Landed as `1b6264d`.

Maintainer verification on exact head `d7ef937`:

- `make test` passed: 374 tests.
- `make lint` passed with 0 serious violations; only 9 existing repository warnings remain.
- `make build` passed for the universal CLI and bridge helper.
- source-blind CLI behavior validation passed for help discoverability, invalid-input fail-closed behavior, and the recorded live bridge contract.
- Codex autoreview findings were fixed for prefixed reply references, NULL plain-message association types, and queued poll sends without a synchronous GUID; final autoreview was clean.
- exact-head macOS and Linux CI passed.
- live contributor proof on a macOS 26 v2 bridge showed the question as a plain caption without reply-thread metadata, native-equivalent database shape, and inbound question backfill through history.

Thanks @omarshahine!

PR #156, opened 2026-07-05
[codex] Include arm64e slice in imsg bridge helper

omarshahine:
## Summary

- build the injected `imsg-bridge-helper.dylib` with an `arm64e` slice by default, while leaving the CLI default at `arm64 x86_64`
- validate every configured helper architecture slice in both local universal builds and the release/notarization path
- update the release packaging test to lock in the new helper architecture default and slice check

## Root Cause

On Apple Silicon/macOS 26, `/System/Applications/Messages.app/Contents/MacOS/Messages` runs as `arm64e`. The released Homebrew helper only had `x86_64 arm64`, so `DYLD_INSERT_LIBRARIES` was set correctly but dyld refused to load the helper before the bridge constructor could run.

Observed failure before the fix:

```text
dyld: inserted dylib ... imsg-bridge-helper.dylib could not be loaded ... fat file, but missing compatible architecture (have x86_64,arm64, need arm64e)
```

## Proof

- `bash -n scripts/build-universal.sh scripts/sign-and-notarize.sh` passed.
- Rebuilt a proof helper with `-arch arm64e -arch arm64 -arch x86_64`; `lipo -info` reported `x86_64 arm64 arm64e`.
- Verified each helper slice explicitly: `arm64e`, `arm64`, and `x86_64`.
- Installed the rebuilt helper locally over the Homebrew 0.12.2 helper and ran `imsg launch --json`; launch succeeded.
- `lsof -p $(pgrep -x Messages)` showed Messages mapped `/opt/homebrew/Cellar/imsg/0.12.2/libexec/imsg-bridge-helper.dylib`.
- `imsg status --json` reported `advanced_features: true`, `v2_ready: true`, and `bridge_version: 2`.

## Test Note

`swift test --filter ReleasePackagingTests` is blocked on this local machine before reaching the packaging test because the active developer directory is Command Line Tools only and test compilation fails with `no such module Testing`. That was pre-existing during the health check; CI/full Xcode should exercise the updated test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 0.12.3 - Unreleased
 
+### Packaging
+- fix: include and validate an arm64e slice in the injected bridge helper for macOS 26 Messages compatibility (#156, thanks @omarshahine).
+
 ## 0.12.2 - 2026-07-04
 
 ### Native Polls
--- a/Tests/imsgTests/ReleasePackagingTests.swift
+++ b/Tests/imsgTests/ReleasePackagingTests.swift
@@ -34,11 +34,18 @@ func universalBuildScriptShipsArm64eHelperSlice() throws {
 func signAndNotarizeScriptDefaultsHelperToArm64e() throws {
   let script = try readRepositoryFile("scripts/sign-and-notarize.sh")
 
-  // The notarize path defaults the helper to arm64e as well, and validates the
-  // shipped slices against the HELPER arch list (not the CLI ARCH_LIST).
+  // The notarize path defaults the helper to arm64e as well, and its lipo guard
+  // must validate the HELPER arch list — not the CLI ARCH_LIST, which omits
+  // arm64e. Assert the loop and its lipo check as one contiguous block so this
+  // can't pass by matching the separate clang-args HELPER_ARCH_LIST loop.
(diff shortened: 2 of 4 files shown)


omarshahine:
Fixed the P1 (64ab396→19c5673): `sign-and-notarize.sh` helper guard now loops `HELPER_ARCH_LIST` instead of `ARCH_LIST`, so arm64e is actually validated in the notarize path. The packaging test now asserts the loop **and** its lipo check as one contiguous block, so it can't pass by matching the separate clang-args `HELPER_ARCH_LIST` loop. macOS + Linux green.

@clawsweeper re-review

steipete:
Landed as `1ee36dd`.

Maintainer verification on exact head `c790b69`:

- `make build` passed; helper reported `x86_64 arm64 arm64e`, every slice verified, relocatable install name and code signatures verified.
- `make test` passed: 368 tests.
- `make lint` passed with 0 serious violations; only the repository's existing warnings remain.
- focused `ReleasePackagingTests` passed: 6 tests.
- Codex autoreview found no actionable issues.
- exact-head macOS, Linux, and Socket checks passed.
- live bridge proof from the contributor: the rebuilt helper loaded into Messages on a SIP-disabled macOS 26 host; `imsg launch` succeeded and status reported advanced features ready, bridge v2 ready, bridge version 2.

Thanks @omarshahine!

Issue #158, opened 2026-07-06
RPC bridge send intermittently delivers one send request twice (~15–30%; two GUIDs, fixed ~60ms gap)


lowcode-garage:
## Summary

A single JSON-RPC `send` request over `imsg rpc --json` intermittently produces **two** physical iMessages in Apple Messages — two distinct `chat.db` rows with different GUIDs at a **fixed ~60–64ms gap** — while the RPC response reports one `ok:true` with a single `guid` (`transport:"bridge"`). Reproduced at ~30% on bare-`to` sends and ~17% on `chat_guid` sends in controlled runs.

## Environment
- imsg **0.11.1** (Homebrew, steipete/tap), bridge transport active (`bridge` helper injected)
- macOS 26.4, Apple Silicon (M4 Mac mini)
- One fresh `imsg rpc --json` process per send (mirrors how OpenClaw's iMessage channel drives sends); no other sender running

## Repro

```bash
for i in $(seq 1 30); do
  printf '{"jsonrpc":"2.0","id":1,"method":"send","params":{"to":"+1XXXXXXXXXX","text":"probe %s","service":"imessage"}}\n' "$i" | imsg rpc --json
  sleep 4
done
# then: chat.db outbound grouped by (chat, second) — count>1 = a physical double
```

## Observed

- 30 bare-`to` sends → **39** chat.db rows: probes 3, 4, 14, 18, 19, 20, 23, 26, 28 each delivered **twice** (~30%).
- 12 `chat_guid` (`any;-;+E164`) sends → **14** rows: 2 doubled (~17%). So both target forms are affected.
- Every double: two **distinct GUIDs**, gap consistently **60–64ms**; the RPC-acknowledged GUID is sometimes the FIRST row of the pair, sometimes the SECOND. Both rows `is_sent=1`, same service/account, identical body. The recipient's device shows the message twice.
- Every RPC response was a single `ok:true` + one guid; all returned GUIDs exist in chat.db — the second row is an extra send the bridge committed, not a client retry (and not an Apple resend, which would reuse the GUID).

## Why it matters

Anything driving sends through the rpc bridge (e.g. OpenClaw's iMessage channel) intermittently double-texts recipients. We chased this for days as an OpenClaw 2026.6.11 dispatch bug before isolating it to bare `imsg rpc` with no orchestrator in the loop (one hook trace + one brain persist + one rpc call → two chat.db rows).

[...]

PR #159, opened 2026-07-06
fix: make bridge rpc text sends idempotent

steipete:
## Summary

- atomically claim each v2 bridge inbox file before reading or dispatching it, making request ownership cross-process instead of process-local
- bootstrap the bridge only inside the `com.apple.MobileSMS` process and keep skipped injected processes from deleting the real bridge-ready lock during teardown
- discard dead/stale claim files without replaying a possibly delivered side effect
- add focused ordering, cleanup, and bootstrap regression coverage plus a 0.12.4 changelog entry

Refs #158.

## Why

The reported failure produces two physical rows with distinct GUIDs from one logical bridge request at a stable ~60–64 ms gap, while the single response can identify either row. That matches two injected inbox watchers reading and dispatching the same finalized request before either removes it.

The previous candidate pinned a client-generated IMMessage GUID. Current macOS 26 plain-text construction already assigns one GUID before its single dispatch, so that approach did not prevent two `sendMessage:` calls. This rewrite fixes ownership at the queue boundary: a same-directory atomic rename gives exactly one process the request before any side effect.

## Verification

- rebased onto `2ea890f` (`main`)
- `DEVELOPER_DIR=/Applications/Xcode-26.6.app/Contents/Developer make test` — 375 tests passed
- `DEVELOPER_DIR=/Applications/Xcode-26.6.app/Contents/Developer make lint` — passed; 7 existing warnings, 0 serious violations
- `DEVELOPER_DIR=/Applications/Xcode-26.6.app/Contents/Developer make build` — universal CLI (`arm64`, `x86_64`) and helper (`arm64e`, `arm64`, `x86_64`) built successfully
- focused helper compilation and claim-ordering test passed after the review fix
- Codex autoreview — first pass found missing orphan-claim cleanup; fixed; local rerun and final branch review both clean with no accepted/actionable findings

## Live validation gate

Not merge-ready yet. The available test host has SIP enabled and reports `v2_ready: false`, so it cannot exercise injected IMCore delivery.

@lowcode-garage please run the issue's fresh-process probe against this head on the reproducing SIP-disabled Mac for both direct `to` and `chat_guid` sends. Please include request count, response count, sender-row count, unique text/GUID count, response-GUID-to-row mapping, duplicate/missing texts, and recipient-side confirmation that each request delivered exactly once. The bridge log now records request UUID, PID, and process name so we can also confirm one handler claimed each request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@
 ### Advanced IMCore
 - feat: add bridge-backed Apple URL preview sends through `send-rich --url` and `send.rich`, with an eight-second out-of-process preparation deadline, capped image decode/staging, and metadata-only fallback (#165, thanks @omarshahine).
 
+### JSON-RPC
+- fix: atomically claim bridge RPC inbox files before dispatch so multiple injected consumers cannot deliver one logical send twice (#158).
+
 ## 0.12.3 - 2026-07-06
 
 ### Native Polls
--- a/Sources/IMsgHelper/IMsgInjected.m
+++ b/Sources/IMsgHelper/IMsgInjected.m
@@ -22,6 +22,7 @@
 #import <pwd.h>
 #import <errno.h>
 #import <stdio.h>
+#import <signal.h>
 #import <sys/stat.h>
 #import <dlfcn.h>
 
@@ -134,11 +135,12 @@ static NSString *kDebugLogFile = nil; // .imsg-bridge.log
 
 static NSTimer *fileWatchTimer = nil;
 static NSTimer *rpcInboxTimer = nil;
-static NSMutableSet *processedRpcIds = nil;
+static BOOL bridgeDidBootstrap = NO;
 static os_unfair_lock eventsLock = OS_UNFAIR_LOCK_INIT;
 static int lockFd = -1;
 
 static const NSUInteger kEventsRotateBytes = 1 * 1024 * 1024;
+static const NSTimeInterval kV2ClaimMaxAge = 10 * 60;
 
 static void initFilePaths(void) {
     if (kCommandFile == nil) {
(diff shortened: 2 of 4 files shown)


steipete:
Landing proof for `f1c74b5239619f99849bb19740fe2aa6fd7b7fc6`:

- Root cause: v2 bridge consumers coordinated only with a process-local set, so multiple injected processes could read and dispatch one finalized inbox request before either removed it.
- Fix: claim each finalized inbox file with a same-directory atomic rename before reading or dispatching; bootstrap only in Messages; remove abandoned claims without replaying a possibly delivered side effect.
- Current-main integration: rebased after the rich-link and poll-unvote bridge changes. Also fixes the newly exposed poll selector validation ordering so mutually exclusive selectors fail before opening `chat.db`.
- Focused proof: bridge claim-ordering test passed; poll conflicting-selector regression test passed; injected helper compiled.
- Full proof: `make test` passed 450 tests; `make lint` passed with 11 existing non-serious warnings; `make build` produced universal CLI (`arm64`, `x86_64`) and helper (`arm64e`, `arm64`, `x86_64`) artifacts.
- Review: final structured review raised the deliberate at-most-once orphan policy. Rejected as intentional: replaying a dead-owner claim can duplicate an already-delivered side effect; solving both crash-window loss and duplication requires a new durable side-effect idempotency protocol outside this fix.
- Live proof: the owner explicitly waived the live Messages bridge requirement for this item after no approved existing SIP-disabled test host was available.
- Public model identifier gate: PASS; exact diff, changed surfaces, workflows, generated metadata, binaries, and public proof text contain no model identifiers.
- Changelog retained because repository policy requires user-visible fixes in the active unreleased section.

Exact-head CI is now the remaining landing gate.

steipete:
Landed on `main` as `0b5dea7efdcab6e906549bd4f854b41f249f8e48`.

Final proof: focused bridge claim and poll selector tests passed; `make test` passed 450 tests; `make lint` passed with existing non-serious warnings; `make build` produced universal CLI and helper artifacts; exact-head macOS/Linux CI passed in https://github.com/openclaw/imsg/actions/runs/29162622981. The owner explicitly waived unavailable live bridge delivery proof for this item. Public model identifier gate: PASS.

PR #160, opened 2026-07-06
Add unread_count and read state for chats/messages

chiedo:
## Summary

Expose read/unread state that already lives in `chat.db` so callers don't have
to re-derive it:

- `chats.list` (RPC) and `imsg chats --json` now include `unread_count` per
  chat — a count of inbound messages where `is_read = 0`.
- `messages.history` (RPC) and `imsg history --json` now include `is_read`
  and `date_read` on inbound messages (omitted for outbound/`is_from_me`
  messages, where the columns aren't meaningful).
- Added `unread_only` (RPC param) / `--unread-only` (CLI flag) to filter
  `chats.list` / `imsg chats` down to chats with `unread_count > 0`.
- All of the above degrade gracefully on older `chat.db` schemas that lack
  `is_read` / `date_read` columns (`unread_count` reports `0`, `is_read` /
  `date_read` are omitted) — detected via the existing `MessageStoreSchema`
  column-probing.

This is read-only: no writes to `chat.db`, no new permissions, no change to
existing output shape for callers that don't look at the new fields.

## Docs

- `docs/rpc.md`: documented `unread_only` param on `chats.list`.

## Test plan

- [x] `make test` — full suite green (372 tests), including new coverage:
  - `Tests/IMsgCoreTests/MessageStoreUnreadStateTests.swift`
  - `Tests/imsgTests/UnreadStatePayloadTests.swift`
- [x] `make build` — release binary builds and runs
- [x] Manually verified against a real `~/Library/Messages/chat.db`:
  `imsg chats --json` and `imsg chats --unread-only --json` both return the
  expected `unread_count` values.

### DIFF
--- a/Sources/IMsgCore/MessageStore+Chats.swift
+++ b/Sources/IMsgCore/MessageStore+Chats.swift
@@ -19,8 +19,10 @@ private struct ListChatsQuery {
   let sql: String
   let bindings: [Binding?]
 
-  init(limit: Int, schema: MessageStoreSchema) {
+  init(limit: Int, unreadOnly: Bool, schema: MessageStoreSchema) {
     let routing = ChatRoutingSelection(schema: schema)
+    let unreadCountColumn = Self.unreadCountColumn(schema: schema)
+    let havingClause = unreadOnly ? "HAVING unread_count > 0" : ""
     if schema.hasChatMessageJoinMessageDateColumn {
       self.sql = """
         SELECT c.ROWID AS chat_rowid, IFNULL(c.display_name, c.chat_identifier) AS name,
@@ -28,10 +30,12 @@ private struct ListChatsQuery {
                MAX(cmj.message_date) AS last_date,
                \(routing.accountIDColumn) AS account_id,
                \(routing.accountLoginColumn) AS account_login,
-               \(routing.lastAddressedHandleColumn) AS last_addressed_handle
+               \(routing.lastAddressedHandleColumn) AS last_addressed_handle,
+               \(unreadCountColumn)
         FROM chat c
         JOIN chat_message_join cmj ON c.ROWID = cmj.chat_id
         GROUP BY c.ROWID
+        \(havingClause)
         ORDER BY last_date DESC
         LIMIT ?
         """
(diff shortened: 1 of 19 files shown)


chiedo:
Pushed two commits addressing the proof/docs asks above:

**`8483fce` — fix: register `--unread-only` as a boolean flag, not a value option**

While generating proof for this review, found that `--unread-only` was actually broken: it was registered as a value-consuming option (`OptionDefinition` in the `options:` array) but read via `values.flag("unread-only")`, which only checks the separate boolean `flags:` collection. So the flag silently did nothing — `imsg chats --unread-only --json` (as suggested above) errored with `Missing value for option unread-only`, and passing an explicit value (`--unread-only true`) still returned the unfiltered list. The RPC path (`chats.list` / `unread_only`) was unaffected, since it parses its own JSON param independently — that's why RPC tests passed while the CLI flag was silently broken.

Fixed by moving `unread-only` into the `flags:` array as a bare boolean flag (matching `HistoryCommand`'s `--attachments` pattern), and added a CLI-level regression test (`ChatsCommandTests.swift`) that drives it through real `argv` parsing via `CommandRouter.run`, not just direct `MessageStore`/RPC calls — a fixture- or RPC-only test wouldn't have caught this, since the bug was purely in the CLI arg wiring.

**`d0d17e5` — docs: document `unread_count`, `is_read`, `date_read`, and `--unread-only`**

Updated `docs/chats.md` and `docs/json.md`, which were still stale (`docs/chats.md` said "`imsg chats` does not take filter flags"; neither doc mentioned the new fields).

**Redacted proof against a real Messages DB, post-fix:**

```
$ imsg chats --json | wc -l
20
$ imsg chats --unread-only --json
{"id": 29, "is_group": false, "service": "SMS", "unread_count": 2}
{"id": 5, "is_group": true, "service": "RCS", "unread_count": 2}
```

`--unread-only` now correctly returns only the chats with unread inbound messages, matching the RPC path's behavior. (Identifiers, names, phone numbers, and account IDs redacted from the full JSON lines; happy to share unredacted with a maintainer directly if needed.)

steipete:
Landed the improved maintainer rewrite in #170: https://github.com/openclaw/imsg/commit/d9bd9c754d50657c5e1b1083d39414ee7ec0e076

The replacement keeps the proposed API, adds schema-safe omission semantics, fixes CLI flag parsing, counts logical URL-preview messages correctly, and adds CLI/RPC/live regression proof. I could not push those fixups to this PR because its branch has maintainer edits disabled.

Thank you @chiedo for proposing the feature and iterating on it. Contributor credit is preserved in the landed commit and changelog. Enabling **Allow edits by maintainers** on future PRs lets us improve the original branch directly.

PR #176, opened 2026-07-12
fix(media): bound ffmpeg conversion wait with a timeout

SebTardif:
## What Problem This Solves

`AttachmentResolver.runConversionProcess` started ffmpeg and blocked on `process.waitUntilExit()` with no deadline. A hung converter stalled attachment metadata resolution indefinitely when unsupported media conversion was enabled.

### Patch

- Bound conversion with a **monotonic** `ContinuousClock` deadline (default 60s).
- On timeout: SIGTERM, then SIGKILL of process and process-group (best-effort).
- Hung regression fixture uses `exec sleep 30` so the Process PID is the sleeper (no orphaned shell child).

### Real ffmpeg success (production conversion args)

CAF generated with ffmpeg, then converted with the same flags `AttachmentResolver` uses (`-nostdin -y -i … -c:a aac -b:a 128k`):

```console
$ ffmpeg -hide_banner -y -f lavfi -i sine=frequency=440:duration=0.5 -c:a pcm_s16le /tmp/imsg-176-proof/tone.caf
$ ffmpeg -hide_banner -nostdin -y -i /tmp/imsg-176-proof/tone.caf -c:a aac -b:a 128k /tmp/imsg-176-proof/tone.m4a
real 0.02
$ file /tmp/imsg-176-proof/tone.m4a
ISO Media, Apple iTunes ALAC/AAC-LC (.M4A) Audio
$ ls -la /tmp/imsg-176-proof/tone.m4a
-rw-r--r--  8544  tone.m4a
```

### Production helper path (AttachmentResolver.metadata + runConversionProcess)

On macOS with Homebrew ffmpeg on PATH, calling the patched `AttachmentResolver.metadata(… convertUnsupported: true)` for the generated CAF:

```console
PROOF real_ffmpeg convertedPath=.../Library/Caches/imsg/converted-attachments/tone-19053d020182e391.m4a
PROOF real_ffmpeg convertedMime=audio/mp4
PROOF hung_status=143 hung_elapsed=0.626815583 seconds
```

`143` is `128 + SIGTERM` from the timeout path; hung `exec sleep 30` returned in under 1s instead of blocking 30s.

### Hung bound (exec sleep fixture)

```console
$ # hung converter: #!/bin/sh + exec sleep 30; timeout 0.5s via production helper
hung_status=143 hung_elapsed≈0.63s process reaped
```

## Real behavior proof

- **Behavior or issue addressed:** Unbounded `waitUntilExit` on ffmpeg conversion could hang attachment metadata forever; conversions are now deadline-bounded and reaped.
- **Real environment tested:** macOS arm64, ffmpeg 8.1.2 (Homebrew `/opt/homebrew/bin/ffmpeg`), imsg branch `fix/ffmpeg-conversion-timeout` at tip after monotonic/process-group fix.
- **Exact steps or command run after this patch:**

  ```bash
  ffmpeg -y -f lavfi -i sine=frequency=440:duration=0.5 -c:a pcm_s16le tone.caf
  ffmpeg -nostdin -y -i tone.caf -c:a aac -b:a 128k tone.m4a
  # plus AttachmentResolver.metadata(convertUnsupported: true) on tone.caf
  # plus runConversionProcess(exec sleep 30, timeout: 0.5)
  ```

- **Evidence after fix:** terminal output above; real m4a produced; hung path returns status 143 in ~0.63s.
- **Observed result after fix:** Successful CAF→m4a conversion via the same args and metadata path; deliberately hung converter exits within the configured bound instead of sleeping 30s.
- **What was not tested:** Live Messages.app chat history with SIP disabled; conversions that legitimately need more than 60s.

## Test plan

- [x] `swift test --filter attachmentResolverConversionTimesOutOnHungConverter`
- [x] `swift test --filter attachmentResolverConversionDoesNotBlockOnConverterOutput`
- [x] Real ffmpeg CAF→m4a on this machine
- [x] Production helper hung path (status 143, sub-second)

### DIFF
--- a/Sources/IMsgCore/AttachmentResolver.swift
+++ b/Sources/IMsgCore/AttachmentResolver.swift
@@ -1,3 +1,4 @@
+import Darwin
 import Foundation
 
 #if canImport(CryptoKit)
@@ -118,7 +119,15 @@ enum AttachmentResolver {
     }
   }
 
-  static func runConversionProcess(executableURL: URL, arguments: [String]) throws -> Int32 {
+  /// Default bound for external converters (ffmpeg). Hung converters must not
+  /// block attachment metadata resolution indefinitely.
+  static let conversionProcessTimeout: TimeInterval = 60
+
+  static func runConversionProcess(
+    executableURL: URL,
+    arguments: [String],
+    timeout: TimeInterval = conversionProcessTimeout
+  ) throws -> Int32 {
     let process = Process()
     process.executableURL = executableURL
     process.arguments = arguments
@@ -126,7 +135,25 @@ enum AttachmentResolver {
     process.standardError = FileHandle(forWritingAtPath: "/dev/null")
 
     try process.run()
-    process.waitUntilExit()
+
+    let deadline = Date().addingTimeInterval(max(0.05, timeout))
+    while process.isRunning {
+      if Date() >= deadline {
+        process.terminate()
+        // Allow a brief grace period for SIGTERM cleanup, then force-kill.
+        let killDeadline = Date().addingTimeInterval(0.5)
+        while process.isRunning, Date() < killDeadline {
+          Thread.sleep(forTimeInterval: 0.02)
+        }
+        if process.isRunning {
+          process.interrupt()
+          kill(process.processIdentifier, SIGKILL)
+        }
+        process.waitUntilExit()
+        return 128 + SIGTERM
+      }
+      Thread.sleep(forTimeInterval: 0.05)
+    }
     return process.terminationStatus
   }
 
--- a/Sources/IMsgCore/AttachmentResolver.swift
+++ b/Sources/IMsgCore/AttachmentResolver.swift
@@ -136,25 +136,40 @@ enum AttachmentResolver {
 
     try process.run()
 
-    let deadline = Date().addingTimeInterval(max(0.05, timeout))
+    // Monotonic deadline so wall-clock jumps cannot stretch the bound.
+    let clock = ContinuousClock()
+    let bound = Duration.seconds(max(0.05, timeout))
+    let deadline = clock.now + bound
     while process.isRunning {
-      if Date() >= deadline {
-        process.terminate()
-        // Allow a brief grace period for SIGTERM cleanup, then force-kill.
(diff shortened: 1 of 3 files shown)


SebTardif:
@clawsweeper re-review

Updated proof: monotonic ContinuousClock deadline, process-group terminate, exec-based hung fixture, and real ffmpeg CAF→m4a terminal output plus production helper hung bound (status 143 in ~0.63s).

steipete:
Maintainer repair pushed in `9b45553` and `8981820`.

The repair moves the timeout regression out of an over-limit test file, strengthens the exact timeout result, and fixes descendant cleanup when the converter leader exits on SIGTERM while a child ignores it.

Exact-head proof at `8981820`:

- `make lint` — pass; 12 pre-existing warnings, 0 serious violations.
- `make test` — pass; 479 tests in 3 suites.
- `make build` — pass; universal `imsg` (`arm64 x86_64`) and helper (`arm64e arm64 x86_64`) built and signed.
- `swift test --filter AttachmentResolverProcessTests` — 2/2 pass.
- Descendant regression repeated 5 additional times with `--skip-build` — 5/5 pass.
- Runtime probe confirmed the launched converter is its process-group leader on this macOS Foundation runtime (`pid == pgid`); the non-`exec` wrapper test confirms its TERM-ignoring descendant is gone before return.

AutoReview:

- Local test-move repair: clean, no accepted/actionable findings.
- Process-group repair after two accepted test/process findings: clean, no accepted/actionable findings.
- Final branch pass repeated a premise contradicted by the runtime probe and 7/7 descendant-test passes (that `Process` inherits the caller group); rejected with evidence. The `getpgid(pid) == pid` guard safely falls back to direct-PID termination if a runtime does not create the group.

Live built CLI attachment-history proof remains blocked on this worker because its parent process lacks Full Disk Access. The contributor's PR body retains real ffmpeg and production-helper proof.

steipete:
Landed with the 60-second converter ceiling accepted as the default: an optional conversion failure is strictly better than an indefinite CLI/RPC hang, the constant is documented and trivially adjustable, and the process-tree cleanup is regression-tested against TERM-ignoring descendants. Changelog entry added on main.

PR #178, opened 2026-07-14
feat(status): report group participant selector availability

markxiong0122:
## Summary
- vend add-participant handles through the existing candidate/service-aware resolver instead of bare `IMHandleWithID:`
- prefer Tahoe selectors `inviteParticipants:reason:` and `removeParticipants:reason:` with the older selectors as fallbacks
- expose selector availability in `status --json`

## Verification
- `make build-dylib`
- `make test` (477 tests)
- live macOS 26 Messages.app gate: add `+13129001215` to a disposable 3-member group, verify via a fresh `imsg chats`, remove it, verify the original participant set is restored

### DIFF
--- a/Sources/IMsgHelper/IMsgInjected.m
+++ b/Sources/IMsgHelper/IMsgInjected.m
@@ -950,6 +950,15 @@ static NSDictionary* handleStatus(NSInteger requestId, NSDictionary *params) {
     NSDictionary *nicknameSelectors = nicknameSharingSelectorStatus();
     Class stickerTransferClass = NSClassFromString(@"IMFileTransfer");
     Class transferCenterClass = NSClassFromString(@"IMFileTransferCenter");
+    Class chatClass = NSClassFromString(@"IMChat");
+    BOOL addParticipant =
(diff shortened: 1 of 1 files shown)


steipete:
Maintainer triage recommendation: **CLOSE**.

The user-visible participant repair this PR originally carried is already on `main` and shipped in v0.13.4: current code uses the service-aware handle resolver and the macOS 26 add/remove selector families. Replaying this stale branch would conflict and regress newer edit-message capability/status work.

The only distinct remainder is two advisory `status --json` booleans. That small addition has no focused coverage and cannot be live-proved on this host: the release binary reports SIP enabled and `v2_ready=false`, so the injected helper cannot expose runtime selector availability without changing the machine’s Messages/SIP setup. The contributor’s stated mutation proof is not inspectable, and it covers behavior that has already shipped rather than the remaining status payload.

Given the shipped substantive fix, dirty branch, regression risk, and weak value of two unproven advisory fields, a status-only rewrite is not worth carrying as a separate PR. This is a close recommendation, not a request for contributor revisions.

PR #179, opened 2026-07-14
fix: use dedicated group photo transfers on macOS 26

markxiong0122:
## Summary
- create group icons with `createNewOutgoingGroupPhotoTransferWithLocalFileURL:` instead of the normal attachment pipeline
- dispatch the returned `at_0_...` transfer GUID through the registry-level group-photo updater with an IMChat fallback
- expose the runtime selector gate in `status --json`

## Root cause
Normal outgoing attachment transfers create a bare UUID and can surface as an image message. Native group icons are a separate transfer type: an orphan `GroupPhotoImage` attachment with an `at_0_...` GUID.

## Verification
- `make build-dylib`
- `make test` (477 tests)
- live macOS 26 Messages.app gate: `chat.properties.groupPhotoGuid` changed to `at_0_...`; no attachment join/filename or visible image bubble was created

### DIFF
--- a/Sources/IMsgHelper/IMsgInjected.m
+++ b/Sources/IMsgHelper/IMsgInjected.m
@@ -429,6 +429,7 @@ static void probeSelectors(void) {
 @interface IMFileTransferCenter : NSObject
 + (instancetype)sharedInstance;
 - (NSString *)guidForNewOutgoingTransferWithLocalURL:(NSURL *)url;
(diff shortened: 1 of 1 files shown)


markxiong0122:
Closing as unverified. The RPC returned success and updated local chat state on macOS 26, but we could not verify receiver-visible group-photo delivery end to end. We may revisit once there is a reproducible visual verification path.

PR #182, opened 2026-07-17
fix: restore Linux attachment resolver build

omarshahine:
## Summary

- restore the Linux read-core build after #176 introduced an unconditional `Darwin` import
- use the same `canImport(Darwin)` / `canImport(Glibc)` platform pattern already used by `SecurePath`

## Reproduction

The two latest `main` workflow runs fail in `linux-read-core` while compiling `AttachmentResolver.swift`:

```text
error: no such module 'Darwin'
```

This also makes unrelated PRs appear red when GitHub tests their merge ref against current `main`.

## Verification

- current `main` failure: https://github.com/openclaw/imsg/actions/runs/29522912373
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer make lint`: exit 0 with 12 existing warnings and 0 serious violations
- `git diff --check`: passed
- macOS build and the affected `AttachmentResolver` process tests compile and run under Xcode; the full local suite reached 477 passing tests before two unrelated rich-link image-staging tests failed
- the PR Linux lane is the authoritative proof for the restored `Glibc` build

### DIFF
--- a/Sources/IMsgCore/AttachmentResolver.swift
+++ b/Sources/IMsgCore/AttachmentResolver.swift
@@ -1,5 +1,10 @@
-import Darwin
 import Foundation
+
+#if canImport(Darwin)
+  import Darwin
+#elseif canImport(Glibc)
+  import Glibc
+#endif
 
 #if canImport(CryptoKit)
   import CryptoKit


steipete:
Maintainer proof on exact head `9ffe9a0a45d5b9609535b1a03ee1376c5b44c836`:

- Provider: brokered AWS Crabbox
- Lease: `cbx_2c219055faf2` (`c7a.8xlarge`, x86_64 Linux)
- Runtime: official `swift:6.2.4-noble` container
- Command path: resolve packages, patch dependencies, `swift test`, `swift build --product imsg`, then `.build/debug/imsg --version`
- Observed: `AttachmentResolver.swift` compiled through the Glibc branch; all 4 Linux read-core tests passed; CLI build passed; real binary printed `0.13.0`.

The lease was released after proof.

@clawsweeper re-review

PR #187, opened 2026-07-20
fix(rpc): do not block startup on undetermined Contacts auth

### LINKED ISSUE #186: rpc: avoid blocking startup on an undetermined Contacts permission
`imsg rpc` initializes Contacts before starting the JSON-RPC server:

- `RpcCommand.swift` calls `await ContactResolver.create()`
- the default policy is `.requestIfNeeded`
- `.notDetermined` then awaits `CNContactStore.requestAccess`

In a headless Aqua-domain LaunchAgent, that authorization request can remain pending indefinitely. The process stays alive but accepts no JSON-RPC requests, so unrelated methods such as `chats.list`, `messages.history`, and `send.rich` all hang before the server starts. We reproduced this on macOS with imsg 0.12.3 and verified the same call path on current `main`.

`ContactResolver` already supports the right fail-open-for-metadata behavior: `.skipIfNotDetermined` returns `NoOpContactResolver(contactsUnavailable: true)`. Could the RPC entrypoint use:

```swift
let contacts = await ContactResolver.create(accessPolicy: .skipIfNotDetermined)
```

Interactive CLI commands can keep the existing default. RPC callers can then start reliably without Contacts display-name resolution, while authorized installations continue loading contacts normally.

A downstream watchdog now bounds requests, but fixing the RPC startup policy avoids the native wedge itself.

SebTardif:
## Update (2026-07-26)

Scoped Contacts fail-open to **noninteractive stdin only** (ClawSweeper P1):
- `stdin` not a TTY → `.skipIfNotDetermined` (headless LaunchAgent/pipe)
- interactive TTY → `.requestIfNeeded` (preserve prompt path)

Tests: `RpcCommandContactsPolicyTests` (headless + interactive).

## What Problem This Solves

`imsg rpc` initializes Contacts **before** starting the JSON-RPC server:

```swift
let contacts = await ContactResolver.create()  // default: .requestIfNeeded
let server = RPCServer(..., contactResolver: contacts)
try await server.run()
```

When Contacts authorization is still `.notDetermined`, `requestIfNeeded` awaits `CNContactStore.requestAccess`. In a headless Aqua-domain LaunchAgent that prompt can remain pending indefinitely. The process stays alive but never accepts JSON-RPC, so methods such as `chats.list`, `messages.history`, and `send.rich` hang before the server starts.

Issue #186 reproduces this on macOS with imsg 0.12.3+ and the same call path on current `main`.

### Prior art (same repo)

- [#136](https://github.com/openclaw/imsg/pull/136) already added `ContactsAccessPolicy.skipIfNotDetermined` and applied it to **`imsg chats`** and **`imsg history`** so CLI list/read commands fail open. **`imsg rpc` was left on the default** when `RpcCommand` was later extracted (commit `b5b7464`, 2026-07-11).
- `ContactResolver.create(accessPolicy: .skipIfNotDetermined)` already returns `NoOpContactResolver(contactsUnavailable: true)` without calling `requestAccess`.

### Fix

Wire RPC startup to the same fail-open policy:

```swift
static let startupContactsAccessPolicy: ContactsAccessPolicy = .skipIfNotDetermined
// ...
await ContactResolver.create(accessPolicy: startupContactsAccessPolicy)
```

Interactive CLI send paths keep the default prompt-if-needed behavior. Authorized installations still load contacts normally on RPC start (existing `ContactResolver` authorized branch; unchanged).

### Live headless transcript (after fix)

Environment: macOS 26.5.2, Contacts authorization **notDetermined** (`CNContactStore.authorizationStatus(for: .contacts) = 0`), built binary at commit `f46cec8`, non-TTY subprocess (LaunchAgent-equivalent headless stdio). Fixture Messages DB via `--db` so startup reaches `ContactResolver.create` (real Messages path would also pass FDA when granted).

```console
$ swift -e 'import Contacts; print(CNContactStore.authorizationStatus(for: .contacts).rawValue)'
0

$ ./bin/imsg rpc --db /tmp/imsg-187-fixture-chat.db
>>> {"jsonrpc":"2.0","id":"p1","method":"chats.list","params":{"limit":5}}
>>> {"jsonrpc":"2.0","id":"p2","method":"messages.history","params":{"chat_id":1,"limit":2}}
<<< {"jsonrpc":"2.0","id":"p1","result":{"chats":[{"id":1,"guid":"iMessage;-;+15555550100","identifier":"+15555550100","name":"Proof Contact","participants":["+15555550100"],"service":"iMessage","is_group":false,"unread_count":0,"last_message_at":"2023-03-08T20:26:40.000Z"}]}}
<<< {"jsonrpc":"2.0","id":"p2","result":{"messages":[{"id":1,"chat_id":1,"text":"hello from fixture","sender":"+15555550100","participants":["+15555550100"],"is_from_me":false,"is_read":true,"created_at":"2023-03-08T20:26:40.000Z",...}]}}
process exit=0 elapsed_s=0.039 responses=2
```

Server accepted two methods in under 50ms while Contacts remained undetermined (no prompt hang). Participant handles left as fixture placeholders (`+15555550100`).

## Real behavior proof

- **Behavior or issue addressed:** Headless `imsg rpc` must start the JSON-RPC server without blocking on an undetermined Contacts permission prompt (hang for LaunchAgents / non-interactive hosts).
- **Real environment tested:** macOS 26.5.2 (Build 25F84), arm64, Contacts authorization status `notDetermined` (raw 0) on the host. Patched binary `f46cec8` at `/tmp/imsg-186/bin/imsg`. Headless non-TTY stdio subprocess (equivalent to LaunchAgent / gateway child). Fixture chat.db so FDA is not required and startup still runs `ContactResolver.create`.
- **Exact steps or command run after this patch:**

  ```bash
  swift -e 'import Contacts; print(CNContactStore.authorizationStatus(for: .contacts).rawValue)'
  # prints 0 (notDetermined)
  make build ARCHES=$(uname -m)
  ./bin/imsg rpc --db /tmp/imsg-187-fixture-chat.db
  # write one JSON-RPC line per request on stdin; read responses on stdout
  ```

- **Evidence after fix:** terminal output from the headless RPC session above: `chats.list` and `messages.history` both returned `result` objects with exit 0 in 0.039s while Contacts stayed undetermined.
- **Observed result after fix:** RPC no longer waits on `requestAccess` when Contacts is undetermined. The server reaches `RPCServer.run()`, accepts methods, and returns JSON-RPC results. Name enrichment remains optional until Contacts is already authorized (fixture used chat `display_name` / handles only).
- **What was not tested:** Live authorized-Contacts path on this host (authorization is notDetermined, not authorized). Existing `ContactResolver` authorized branch is unchanged; interactive CLI send paths still use `.requestIfNeeded`. Real LaunchAgent plist install was not used; proof used an equivalent non-TTY stdio child process.

## Notes

- Closes #186
- Changelog entry under `0.13.2 - Unreleased` / JSON-RPC
- Refactors `RpcCommand` to a `run(..., contactResolverFactory:)` entry point consistent with `ChatsCommand` for testability
- Policy note for maintainers: this extends the fail-open Contacts policy from #136 to RPC startup so headless hosts are reliable by default


SebTardif:
@clawsweeper re-review

Added live headless `imsg rpc` request/response transcript with Contacts authorization `notDetermined` (host status raw 0). Patched binary accepted `chats.list` and `messages.history` with JSON-RPC results in 0.039s (no Contacts prompt hang). Fixture DB via `--db` so startup reaches ContactResolver; non-TTY stdio child is LaunchAgent-equivalent. PR body Evidence + Real behavior proof sections updated.

SebTardif:
The ClawSweeper review raises a fair point: making `skipIfNotDetermined` unconditional for all RPC changes the contract that PR #136 set up (fail-open for `chats`/`history` reads, but prompt-capable for RPC).

Two options:

1. **Scope to headless** (ClawSweeper recommended): apply `skipIfNotDetermined` only when the RPC startup detects a noninteractive/headless context, and keep the prompt path for interactive sessions. Add coverage for both modes.
2. **Accept all-RPC fail-open**: document that RPC no longer requests Contacts authorization at startup, which means contact-name resolution may be unavailable until the user grants access through another command.

@cemendes, since you introduced the Contacts policy split in #136, could you weigh in on which approach fits the intended design? Happy to implement whichever direction you prefer.

SebTardif:
@clawsweeper re-review

Scoped fail-open Contacts policy to noninteractive stdin (isatty). Interactive TTY keeps requestIfNeeded. Policy unit tests added for both modes.

SebTardif:
Linux CI regression fixed: unconditional `import Darwin` broke `linux-read-core`. Now uses the same `#if canImport(Darwin) / Glibc` pattern as `SecurePath`.

`linux-read-core` and `macos` are green on head `b8bcf52`.

steipete:
Maintainer repair is now on the branch at `af95a7c`.

The original startup policy is preserved: non-interactive RPC startup no longer blocks on an undetermined Contacts permission, while interactive startup can still request access. I also closed the unsafe fallback that review exposed: when Contacts are unavailable, a name-like target now returns `-32602` before the send closure is invoked; explicit phone numbers, email addresses, and chat IDs remain usable.

Proof on macOS arm64:

- Focused Contacts/RPC suite: 7 tests passed.
- `make lint`: passed (12 pre-existing warnings).
- `make test`: 488 tests passed. The first run hit the existing process-descendant PID-file flake; its exact retry and the complete retry both passed.
- `make build ARCHES="$(uname -m)"`: release CLI and universal helper built successfully.
- Live non-TTY release run with macOS Contacts authorization raw value `0` (not determined): `chats.list` against a temporary Messages-compatible SQLite fixture returned one chat and no RPC error. No message was sent.
- Independent local-diff and full-branch autoreviews: clean, with no actionable findings.

GitHub CI for this exact commit is still running; I’ll leave the final check state in the triage report.

PR #189, opened 2026-07-21
perf(history): fetch reactions per-chat via index instead of full-table scan

### LINKED ISSUE #188: history --json: reaction fetch full-scans the message table per batch (non-indexed LIKE)
### What

`imsg history --json` on a busy chat spends most of its time attaching
reactions. `MessageStore.reactions(for messages:)` matches reactions to their
target with:

```sql
... OR r.associated_message_guid LIKE '%/' || ?
```

The leading wildcard makes that predicate non-sargable, so the
`associated_message_guid` index can't be used. `EXPLAIN QUERY PLAN` for one
200-guid batch:

```
SCAN r
```

i.e. every batch full-scans the entire `message` table. A large chat runs dozens
of full scans (guids / 200) just to attach tapbacks.

### Impact

Measured on a ~197k-message `chat.db` (debug build, best of 3):

| chat | msgs | `history --json` |
|-----:|-----:|-----------------:|
| 5,473 | 5,473 | 2.03s |
| 13,252 | 13,252 | 4.92s |
| 20,755 | 20,755 | 7.54s |

Isolating just the reaction fetch, a 20k-message chat spends ~3.2s of that in
the scan.

### Proposed fix

Fetch the chat's reaction rows through the indexed `chat_message_join` in one
bounded query and match them to their target message in memory. A reaction
always lives in the same chat as the message it targets, so results are
identical. Plan becomes:

```
SEARCH cmj USING COVERING INDEX (chat_id=?)
SEARCH r USING INTEGER PRIMARY KEY (rowid=?)
```

End-to-end this makes `history --json` ~1.7–1.9× faster (e.g. the 20k chat:
7.54s → 4.38s), with byte-identical output.

I have a patch ready (reuses the existing add/remove replay) and will open a PR
referencing this issue.

### Environment

- imsg `main`
- macOS, real Messages `chat.db`

zachwinter:
Closes #188.

## Summary

`reactions(for messages:)` matches reactions to their target with
`associated_message_guid LIKE '%/' || ?`. The leading wildcard makes that
predicate non-sargable, so `EXPLAIN QUERY PLAN` reports `SCAN r` — every
200-guid batch full-scans the entire `message` table. A large chat runs dozens
of full scans just to attach tapbacks, which dominates `history --json`.

This PR adds `reactions(forChatID:messages:)`, which pulls the chat's reaction
rows through the indexed `chat_message_join` in **one bounded query** and
matches them to their target message in memory. Every reaction lives in the same
chat as the message it targets, so the result is identical to the per-message
path. The existing add/remove replay (`applyBulkReactionRow`, `BulkReactionKey`,
`baseAssociatedMessageGUID`) is reused unchanged. `history --json` and the RPC
`messages.history` handler (both single-chat) switch to it; nothing else changes.

## Query plan

```
before (per 200-guid batch):   SCAN r
after  (chat-scoped):          SEARCH cmj USING COVERING INDEX (chat_id=?)
                               SEARCH r USING INTEGER PRIMARY KEY (rowid=?)
```

## Benchmark

Real `history --json` on a 197k-message `chat.db`, debug build, best of 3,
**output byte-identical before/after**:

| chat | msgs | main | this PR | speedup | saved |
|-----:|-----:|-----:|--------:|--------:|------:|
| 81   | 1,776 | 0.60s | 0.33s | 1.80× | 0.27s |
| 50   | 3,556 | 1.30s | 0.76s | 1.71× | 0.54s |
| 1138 | 6,269 | 1.93s | 1.00s | 1.93× | 0.93s |
| 23   | 5,473 | 2.03s | 1.20s | 1.69× | 0.83s |
| 11   | 13,252 | 4.92s | 2.94s | 1.67× | 1.98s |
| 426  | 20,755 | 7.54s | 4.38s | 1.72× | 3.16s |

The isolated reaction fetch itself drops ~200–600× (e.g. chat 426: 3.25s → 5ms);
end-to-end `history` is ~1.7–1.9× faster because reactions are ~40% of the
command.

## Verify

```bash
# pick a busy chat id from `imsg chats`
imsg history --chat-id <ID> --limit 100000 --json > /tmp/after.jsonl
# vs a build of main -> byte-identical (modulo per-process JSON key order)
```

## Testing

- `make test` → **484 tests pass**, including 4 new cases in
  `MessageStoreReactionsTests.swift`: parity vs `reactions(for:)` across
  tapback / custom-emoji / add-then-remove, empty input, missing reaction
  columns, and a 600-message scale check.
- `swift format` lint clean on touched files.
- No new dependencies; the Linux read-core build is unaffected.

## Notes / scope

- `reactions(for messages:)` is left in place (no caller change beyond the two
  single-chat sites); it can be removed in a follow-up if desired.
- Out of scope: a separate, non-reaction bottleneck shows up on attachment-heavy
  chats (e.g. chat 169 spends ~5s outside reactions) — happy to look at that
  next if it's of interest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Performance
 - perf: avoid a temporary SQLite sort for every URL-preview lookup in history (#191, thanks @zachwinter).
+- perf: fetch history reactions in one pass while preserving cross-chat reaction associations (#189, thanks @zachwinter).
 
 ## 0.13.1 - 2026-07-17
 
--- a/Sources/IMsgCore/MessageStore+HistoryMetadata.swift
+++ b/Sources/IMsgCore/MessageStore+HistoryMetadata.swift
@@ -3,7 +3,6 @@ import SQLite
 
 extension MessageStore {
   private static let bulkAttachmentBatchSize = 500
-  private static let bulkReactionBatchSize = 200
 
   public func attachments(
     for messageIDs: [Int64],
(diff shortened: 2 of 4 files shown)


steipete:
Reworked this on the PR branch before landing. Real Messages data can place a reaction row in a different chat, or with no chat join, from its target; the original chat-scoped fetch therefore omitted valid reactions. The replacement scans associated reaction rows once, preserves GUID matching in memory, and bulk-prefetches JSON-RPC history too. It keeps the performance goal while preserving existing output semantics.

PR #195, opened 2026-07-23
Add --redact-codes flag to strip texted security codes

thosss:
## Summary
- Adds a global `--redact-codes` flag (same design as `--read-only`: accepted before or after the subcommand, combinable) that redacts texted security/verification codes (2FA, OTP) from message text wherever it's shown — `history`, `search`, `watch`, `scheduled`, and the equivalent JSON-RPC methods (`messages.history`, `watch.subscribe`, `messages.scheduled`).
- The matching heuristic was derived empirically: mined 218 real OTP/PIN messages from a live chat.db (keyword search + a full sweep of every SMS short-code sender) and cross-checked against 1,589 ordinary messages to measure false positives before writing the regex. Full writeup of findings is in the commit message.
- Notable design points: matches digits before *or* after the keyword (the "123456 is your code" autofill format is at least as common as "code: 123456"), allows dash-formatted codes (WhatsApp/Cash App), intentionally limited to digit/dash tokens (alphanumeric codes are an accepted rare miss), excludes URL-embedded false positives, and is not restricted to short-code senders (Uber/Toast send real 2FA codes from ordinary phone numbers).
- Fixed a real bug found during manual smoke testing: `--redact-codes` before the subcommand broke argument parsing entirely (only `--read-only` had a pre-subcommand scan). Generalized `CommandRouter`'s flag pre-scan to cover both flags, with regression tests.

## Test plan
- [x] `swift build` clean
- [x] `swift test` — 531/531 passing
- [x] `swift format lint` clean
- [x] Manual smoke test against the real local chat.db in `--read-only` mode confirming correct redaction (Marcus, OpenTable, Ring, Apple, Premise Health, Amex, Walmart driver passcode, etc.) with no observed false positives in a live sample
- [x] Regression test for the `--redact-codes`-before-subcommand parsing bug

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Safety
 - feat: add a global `--read-only` flag (and `IMSG_READ_ONLY=1` environment variable) that deterministically refuses every write or mutation across the CLI and JSON-RPC. Read commands are unaffected; write commands exit with a dedicated code (3) and a clear message, and mutating RPC methods return a well-formed JSON-RPC error (`code: -32001`) without breaking the protocol. `imsg status` reports the active mode (`read_only` in `--json`).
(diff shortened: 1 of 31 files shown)


thosss:
Sorry, wrong repo by mistake.

PR #198, opened 2026-07-27
fix(polls): render full selection snapshots [AI-assisted]

clawSean:
## Summary

- render native poll vote rows as full selection snapshots in plain
  `history`/`watch` output
- stop describing the first snapshot entry as the option that just changed
- document `poll.votes` as authoritative and `poll.vote` as a compatibility
  projection
- cover multi-select, one-option-remaining, and empty-selection snapshots

## Why

Apple's native poll vote rows carry a participant's full selected-option set,
not a single delta. `pollDisplayText` currently renders only `poll.vote`, which
is documented as the first decoded entry. Selecting a second option therefore
prints only the first option; removing one of two options can falsely describe
the remaining option as a new vote.

The JSON schema remains compatible. This change only makes human-readable
output follow the already-exposed `poll.votes` snapshot.

## Evidence

The regression failed on current `main` with:

```text
[poll vote] +15550002000 selected Lobster
```

for a snapshot containing both `Lobster` and `Also lobster`.

After the change:

```text
[poll selections] +15550002000 selected Lobster / Also lobster
```

Verification on macOS 26:

- full `swift test`: 488 tests passed
- `swift format lint --recursive Sources Tests TestsLinux`: passed
- `git diff --check`: passed
- `swiftlint` was unavailable on the test host; CI remains authoritative for
  that lane

AI-assisted with Codex. I reviewed the resulting behavior and tests.
### Live native-poll proof (2026-07-28)

Verified on the submitted head `62676f5c9da471018b12b32e44fa91857f745025` against a real native Messages poll on macOS 26.5.1. The participant identifier was redacted before posting.

After selecting a second option in the Messages UI, the patched CLI rendered:

```text
[poll selections] participant-1 selected Beta / Alpha
```

After deselecting Alpha, leaving Beta selected, it rendered:

```text
[poll selections] participant-1 selected Beta
```

This is the key negative case: the remaining Beta selection is described as the current snapshot, not as a new Beta vote.

Current-head validation:

- `swift test`: 488 tests in 3 suites passed
- `swift format lint --recursive Sources Tests TestsLinux`: passed
- `git diff --check origin/main...HEAD`: passed

The earlier automated compile concern does not reproduce. `MessagePollVote.optionID` is a non-optional `String`; only `optionText` is optional, so `vote.optionText ?? vote.optionID` is a concrete `String` and the full current-head test build succeeds.

### DIFF
--- a/Sources/imsg/AttachmentDisplay.swift
+++ b/Sources/imsg/AttachmentDisplay.swift
@@ -49,12 +49,12 @@ func pollDisplayText(for poll: MessagePollEvent) -> String {
         return "\(participant) selected \(selections.joined(separator: " / "))"
       }
       if !snapshots.isEmpty {
-        return "[poll selections] \(snapshots.joined(separator: "; "))"
+        return "[poll vote] \(snapshots.joined(separator: "; "))"
       }
-      return "[poll selections] no options selected"
+      return "[poll vote] no options selected"
     }
     guard let vote = poll.vote else {
-      return "[poll selections] no options selected"
+      return "[poll vote] no options selected"
     }
     let participant = vote.participant ?? "someone"
     let option = vote.optionText ?? vote.optionID
--- a/Sources/imsg/AttachmentDisplay.swift
+++ b/Sources/imsg/AttachmentDisplay.swift
@@ -28,9 +28,37 @@ func pollDisplayText(for poll: MessagePollEvent) -> String {
     let options = poll.options?.map(\.text).joined(separator: " / ") ?? ""
     return options.isEmpty ? "[poll created] \(question)" : "[poll created] \(question): \(options)"
   case .vote:
-    let participant = poll.vote?.participant ?? "someone"
-    let option = poll.vote?.optionText ?? poll.vote?.optionID ?? "unknown option"
-    let action = poll.vote?.eventType ?? "selected"
+    if let votes = poll.votes {
+      var participantOrder: [String] = []
+      var selectionsByParticipant: [String: [String]] = [:]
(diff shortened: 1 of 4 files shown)


clawSean:
Updated the PR body with redacted real-behavior proof from the current head.

The live Messages UI produced a two-selection snapshot, then a one-selection snapshot after deselection. Head `62676f5c9da471018b12b32e44fa91857f745025` rendered both correctly, and the full Swift suite passed 488/488 tests. The prior compile concern does not reproduce because `MessagePollVote.optionID` is non-optional.

@clawsweeper re-review

steipete:
Maintainer repair is on the branch at `bd539a5`; I recommend landing this PR once CI finishes.

The functional fix remains the contributor’s: plain output now renders the authoritative `poll.votes` selection snapshot instead of the compatibility-only first `poll.vote`. I adjusted the final presentation to retain the established `[poll vote]` prefix, avoiding an unnecessary output compatibility break while still showing every selected option.

Proof on macOS arm64:

- Focused utility tests passed, including two simultaneous selections, the remaining-selection case, and an empty snapshot.
- `make lint`: passed with pre-existing warnings only.
- `make test`: all 488 tests passed.
- `make build ARCHES="$(uname -m)"`: release CLI and universal helper built successfully; live `./bin/imsg --version` reported 0.13.5.
- The contributor’s native Messages poll reproduction on this PR demonstrates the bug and the corrected multi-selection output against a real local chat database. The maintainer follow-up changes only the stable prefix.
- Local-diff and full-branch autoreviews: clean, no actionable findings.

No message or poll mutation was performed during maintainer verification.

PR #199, opened 2026-07-28
fix(bridge): fail fast when a request is discarded by a Messages relaunch

omarshahine:
## What Problem This Solves

Fixes an issue where a bridge request in flight when Messages.app restarts blocks the caller for the full timeout instead of failing. For sends that is 150s of a completely unresponsive caller, ending in a timeout for a request that was actually discarded seconds in.

Seen in production on an OpenClaw gateway: the agent's iMessage reply blocked for 150s and surfaced as a failure, even though the bridge had been repaired ~23s in and was healthy again long before the timeout expired.

## Why This Change Was Made

`MessagesLauncher.launchInjectedMessages()` calls `cleanQueueDirectory()` on both queue directories, so relaunching Messages.app with the dylib wipes any request already in flight. When Messages.app dies mid-request and the keepalive relaunches it, the original request file is deleted and nothing will ever write its response. `invokeV2` had no way to notice, so it polled the outbox until the deadline for a reply with no writer.

The request's own on-disk state is a precise signal. A live request is either unclaimed (`<id>.json`) or claimed by the dylib (`<id>.processing.<pid>`, set by `processV2InboxFile`). Neither present, with no reply, means the queue was cleared.

Two details worth review:

- **Race with reply-writing.** The dylib removes its claim and writes the reply as separate steps, so a reply can land between the two checks. The outbox is re-checked once before giving up.
- **Fail-safe direction.** An inbox that cannot be enumerated returns "still queued" so a transient read error never aborts a live request. The change can only end a wait early when the request is provably absent.

Beyond ending the stall, this separates "discarded, definitely not delivered" (`.bridgeNotReady`) from a plain timeout, where delivery is genuinely unknown. That distinction is what lets a caller retry safely: retrying after a timeout risks a duplicate, retrying a discarded request cannot.

## User Impact

A send issued while Messages.app is restarting now fails in about a poll interval instead of blocking for 150s, and reports a cause the caller can act on. Callers that today surface a hard failure to the user can distinguish a discarded request and retry it once the bridge is back.

## Evidence

**Pre-fix production incident.** Timeline reconstructed from macOS unified logs and the gateway log:

| Time (local) | Event |
|---|---|
| 21:16:29 | Messages.app (pid 7836) alive, serving the bridge |
| 21:16:58.587 | `loginwindow … appDeath for com.apple.MobileSMS` — Messages.app dies |
| 21:17:45 | `send-message` issued; request written to the inbox |
| 21:17:42Z | keepalive logs `private bridge unavailable; reinjecting` |
| 21:17:54.807 | Messages.app relaunches as pid 31647 (cold start: `AppSandbox`, `container_create_or_lookup`) — queue directories wiped here |
| 21:18:05Z | keepalive logs `private bridge ready` |
| 21:20:16 | caller finally errors, `Timed out waiting for response to 'send-message'` |

The bridge was healthy from 21:18:05, yet the request waited another ~2 minutes for a reply that could not arrive because its request file no longer existed. That is the window this change closes.

**Tests** (`Tests/IMsgCoreTests/IMsgBridgeClientQueueTests.swift`) cover the three on-disk shapes the loop distinguishes, plus the fail-safe:

- unclaimed `<id>.json` → still queued
- claimed `<id>.processing.<pid>` → still queued (a live request must not be aborted)
- neither, including an unrelated request present → not queued
- unreadable inbox → still queued

`swift test`: 489 tests in 4 suites, all passing. `swift format lint` clean on both touched files; `swiftlint` reports no violations in them; `git diff --check` clean.

**Proof gap, stated plainly:** the exact-path runtime evidence is the pre-fix incident. I have not killed Messages.app mid-send against the patched build — doing that on the production gateway would interrupt a live messaging channel. The discard path is covered deterministically by the tests above.

AI-assisted.

---

## After-fix real behavior proof (added)

Run on macOS 26 against the real injected Messages.app, with the production gateway stopped first so its `imsg rpc` released the bridge and the RPC queue had no competing client (verified `imsg rpc` count 0, inbox empty before the run).

**Fault injection.** `SIGSTOP` on the injected Messages process (verified state `T`) so the dylib physically cannot drain the inbox, then the request is issued, then the process is killed and `imsg launch` relaunches it — which is what invokes `cleanQueueDirectory()` and wipes the in-flight request.

**Early failure — the changed path, patched build:**

```
request queued: EA6D821C-2857-40DD-8BB7-8B99F80D93A2.json
elapsed=11.658s
errorCase=imsg bridge not ready: request for 'typing' was discarded before it was
          processed (Messages.app restarted or the bridge queue was cleared)
```

The call was made with `timeout: 150`. Before this change it would have polled the full 150s and then reported `Timed out waiting for response to 'send-message'`; instead it returns `.bridgeNotReady` as soon as the request is observed to be gone. The elapsed time is dominated by client startup before the request is written; the wait ends promptly once the queue is wiped.

**Recovery — same call after the bridge is back:**

```
elapsed=0.067s
errorCase=Dylib error: Missing required parameter: handle
```

The dylib answers in 67ms with its own parameter validation, confirming the bridge recovered and that the new queue-state check does not produce a false positive against a live bridge.

These were produced with a small throwaway executable calling `IMsgBridgeClient.invoke` directly, because each CLI subcommand wraps bridge errors in its own message (`imsg send` falls back to AppleScript and surfaces `Connection is invalid (-609)`; `imsg typing` surfaces an imagent error), which hides the underlying `IMsgBridgeError`. The probe is not part of this PR — the branch contains only the two files above.

**Supporting CLI timings from the same isolated setup**, showing the wait ending rather than running to the deadline: `imsg send` returned 0.52s after its request was queued, `imsg typing` 0.415s, and a post-recovery send succeeded in 0.64s.

**Environment restored:** production LaunchAgent bootstrapped, listener on :18789, exactly one `imsg rpc`, gateway health `ok`, iMessage channel `running` with `lastError: null` and `reconnectAttempts: 0`, and a live round-trip send verified.

**Remaining honesty note:** the crash that starts this sequence is not addressed here. On the affected machine Messages.app dies on its own roughly every 2-3 days (6 message-tool stalls across 14 days of gateway logs); this change only stops an in-flight request from waiting out the full timeout when that happens.

### DIFF
--- a/Sources/IMsgCore/IMsgBridgeClient.swift
+++ b/Sources/IMsgCore/IMsgBridgeClient.swift
@@ -95,10 +95,6 @@ public final class IMsgBridgeClient: @unchecked Sendable {
     try FileManager.default.moveItem(atPath: tmp, toPath: final)
 
     let deadline = Date().addingTimeInterval(timeout)
-    // Whether the dylib ever claimed this request. It decides which error a
-    // vanished request produces, because the two cases differ in whether the
-    // action can already have run.
-    var wasClaimed = false
     while Date() < deadline {
       try await Task.sleep(nanoseconds: UInt64(pollInterval * 1_000_000_000))
       if let response = try readV2Response(outPath: outPath) {
@@ -117,7 +113,6 @@ public final class IMsgBridgeClient: @unchecked Sendable {
       case .unclaimed:
         continue
       case .claimed:
-        wasClaimed = true
         continue
       case .absent:
         // Re-check the outbox once: the reply is renamed into place before the
@@ -125,19 +120,12 @@ public final class IMsgBridgeClient: @unchecked Sendable {
         if let response = try readV2Response(outPath: outPath) {
           return try unwrapV2Response(response)
         }
-        // Never claimed: nothing read the request, so the action did not run
-        // and the caller can safely retry.
-        guard wasClaimed else {
-          throw IMsgBridgeError.bridgeNotReady(
-            "request for '\(action.rawValue)' was discarded before it was processed "
-              + "(Messages.app restarted or the bridge queue was cleared)"
-          )
-        }
-        // Claimed and then vanished with no reply. processV2InboxFile renames
-        // the reply into the outbox before removing the claim, so this means
-        // the dylib died mid-request: the action may already have taken
-        // effect. Report the outcome as unknown so nothing retries it blind.
-        throw IMsgBridgeError.deliveryUnknown(action: action.rawValue)
+        // A claim can be created, acted on, and removed between two polls. An
+        // absent request therefore never proves that the action did not run,
+        // even when this client did not observe the claimed state. Use the
+        // existing timeout case so callers cannot mistake this for a
+        // retry-safe bridge-not-ready failure.
+        throw IMsgBridgeError.timeout(action: action.rawValue)
       }
     }
 
--- a/Sources/IMsgCore/IMsgBridgeClient.swift
+++ b/Sources/IMsgCore/IMsgBridgeClient.swift
@@ -95,6 +95,10 @@ public final class IMsgBridgeClient: @unchecked Sendable {
     try FileManager.default.moveItem(atPath: tmp, toPath: final)
 
     let deadline = Date().addingTimeInterval(timeout)
+    // Whether the dylib ever claimed this request. It decides which error a
+    // vanished request produces, because the two cases differ in whether the
+    // action can already have run.
+    var wasClaimed = false
     while Date() < deadline {
       try await Task.sleep(nanoseconds: UInt64(pollInterval * 1_000_000_000))
       if let response = try readV2Response(outPath: outPath) {
@@ -109,16 +113,31 @@ public final class IMsgBridgeClient: @unchecked Sendable {
       //
       // A request in normal flight is still on disk: unclaimed as
       // `<id>.json`, or claimed by the dylib as `<id>.processing.<pid>`.
-      if !requestStillQueued(inboxDir: inboxDir, id: id) {
-        // Re-check the outbox once: the dylib removes its claim and writes the
-        // reply as two separate steps, so a reply may have landed in between.
+      switch requestQueueState(inboxDir: inboxDir, id: id) {
+      case .unclaimed:
+        continue
+      case .claimed:
+        wasClaimed = true
+        continue
+      case .absent:
+        // Re-check the outbox once: the reply is renamed into place before the
+        // claim is dropped, so a reply may have landed between our two checks.
         if let response = try readV2Response(outPath: outPath) {
           return try unwrapV2Response(response)
         }
-        throw IMsgBridgeError.bridgeNotReady(
-          "request for '\(action.rawValue)' was discarded before it was processed "
-            + "(Messages.app restarted or the bridge queue was cleared)"
-        )
+        // Never claimed: nothing read the request, so the action did not run
+        // and the caller can safely retry.
+        guard wasClaimed else {
+          throw IMsgBridgeError.bridgeNotReady(
+            "request for '\(action.rawValue)' was discarded before it was processed "
+              + "(Messages.app restarted or the bridge queue was cleared)"
+          )
+        }
+        // Claimed and then vanished with no reply. processV2InboxFile renames
+        // the reply into the outbox before removing the claim, so this means
+        // the dylib died mid-request: the action may already have taken
+        // effect. Report the outcome as unknown so nothing retries it blind.
+        throw IMsgBridgeError.deliveryUnknown(action: action.rawValue)
       }
     }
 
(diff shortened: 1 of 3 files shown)


omarshahine:
@clawsweeper review

omarshahine:
@clawsweeper re-review

omarshahine:
Thanks — the finding pointed at a real problem, though not via the mechanism described. Fixed in `1f87ec9`.

**On the stated ordering.** The review says the producer "removes its processing claim and writes its reply as separate steps, leaving an interval where the action may already have run but the reply is not yet visible." The order is the opposite. In `processV2InboxFile` the reply is renamed into the outbox *before* the claim is dropped:

```objc
response = processV2Envelope(envelope);      // action executes
...
[responseData writeToFile:tmp atomically:NO];
rename(tmp.UTF8String, outPath.UTF8String);  // reply published (atomic)

// Drop the claimed request — we're done with it. If the process dies
// after claiming, a later inbox scan removes the orphan without
// replaying a potentially delivered side effect.
[[NSFileManager defaultManager] removeItemAtPath:claimPath error:nil];
```

So "keep the processing claim until the reply is published" is already the invariant, and reordering would be a no-op.

**But the concern is valid through another path**, which that comment names: if the dylib dies *after* IMCore delivered and *before* the reply is written, the claim is orphaned, and `cleanupOrphanedV2Claims` or a relaunch clears it. My check then saw nothing and reported `.bridgeNotReady`, which this PR documents as retry-safe — for a message that may well have gone out. On the machine this was found on, Messages.app dies every 2-3 days, so that window is not hypothetical.

The defect was my **error semantics**, not the producer ordering: I collapsed two different situations into one retry-safe error.

**Fix.** Track whether the dylib ever claimed the request:

| Observed | Meaning | Error |
|---|---|---|
| never claimed → absent | nothing read it, action did not run | `.bridgeNotReady` (retry-safe, unchanged) |
| claimed → absent, no reply | dylib died mid-request, action may have run | **`.deliveryUnknown(action:)`** — must not be retried blind |

`requestStillQueued` became `requestQueueState` returning `unclaimed` / `claimed` / `absent`. An inbox that cannot be enumerated still reports `unclaimed`, so a transient read error never ends a live request.

Added the interleaving test that was asked for: one request walked through `unclaimed → absent`, another through `unclaimed → claimed → absent`, plus a guard that the two error cases cannot be conflated.

`swift test`: 491 tests in 4 suites passing. `swift format lint` and `swiftlint` clean on all three touched files; `git diff --check` clean.

Two things I did not change, deliberately:

- **No producer-side reordering**, since the invariant already holds. If you would still prefer a durable producer-side record (e.g. keeping the claim until the reply is confirmed readable), say so and I will do it — but it looked like added protocol surface for an invariant that is already satisfied.
- **The residual window is narrowed, not closed.** A crash between delivery and reply publication is still unobservable from the client; the change makes that case report unknown rather than retry-safe, which is the honest outcome rather than a guarantee.

omarshahine:
@clawsweeper re-review

omarshahine:
@clawsweeper re-review

omarshahine:
Note for anyone reading this PR: **the review shown above is stale.** It was produced against `94b5811` and predates the fix for its own P1.

The P1 ("Keep the processing claim until the reply is published") was addressed in `1f87ec9`, along with the interleaving test it asked for. Summary of that exchange, in case the bot verdict never refreshes:

- The finding's stated mechanism is inverted. `processV2InboxFile` renames the reply into the outbox *before* dropping the claim, so the requested ordering is already the invariant and reordering would be a no-op. Quoted source is in [the reply above](https://github.com/openclaw/imsg/pull/199#issuecomment-5108567854).
- The underlying concern is real via a different path, which that function's own comment names: if the dylib dies after IMCore delivered but before the reply is written, the claim is orphaned and later cleared. My original code reported that as retry-safe `.bridgeNotReady`.
- Fixed by splitting the error semantics rather than the ordering: never-claimed-then-absent stays `.bridgeNotReady` (retry-safe); claimed-then-absent-with-no-reply becomes the new `.deliveryUnknown(action:)`, which must not be retried blind.

State at `1f87ec9`: `swift test` 491 tests in 4 suites passing, `swift format lint` and `swiftlint` clean on all touched files, `git diff --check` clean. Proof of the original stall fix was accepted in the last review (`proof: sufficient`).

**Bot-side observation**, in case it is useful to whoever maintains ClawSweeper: re-review requests are acking within seconds but not producing a review, and it only started after the head moved. Reviews ran fine at `94b5811` (ack 06:32 → verdict 06:38; ack 14:50 → verdict 15:23), and all three requests since the push to `1f87ec9` have acked and gone quiet (19:08, 22:15, plus the automatic trigger on the body update). Not re-issuing the command to avoid piling on a possibly-active lease.

steipete:
Maintainer repair is on the branch at `5c63215`; I recommend landing this PR once CI is green.

The fast-failure mechanism is sound, but the original branch inferred retry safety from whether this client happened to observe a `.processing` file. That state can be created, acted on, and removed entirely between two 50ms polls, so “never observed claimed” did not prove the send was safe to retry. The repair treats every vanished request without a reply as delivery-unknown via the existing timeout case, preserving source compatibility and preventing blind duplicate sends.

Proof on macOS arm64:

- Queue-state suite: all 5 tests passed, covering unclaimed, claimed, absent, unreadable, and both absent-state histories.
- `make lint`: passed with pre-existing warnings only.
- `make test`: all 490 tests passed after the known attachment-descendant PID-file flake failed once; its exact retry and the complete retry passed.
- `make build ARCHES="$(uname -m)"`: release CLI and universal helper built successfully.
- Live release status run succeeded and reported 0.13.5, SIP enabled, `v2_ready=false`, and basic features available. That host state makes a second destructive Messages relaunch/fault-injection run inappropriate; the contributor’s real bridge restart reproduction on this PR remains the live proof of the fast-failure path. The maintainer repair changes only retry classification after that same disappearance.
- Local-diff and full-branch autoreviews: clean, no actionable findings.

No message was sent and Messages.app was not relaunched during maintainer verification.

PR #212, opened 2026-08-09
fix: open sticker staging inside Messages sandbox

### LINKED ISSUE #211: send.sticker fails with "Could not securely open sticker directory" inside sandboxed Messages, while the rich-link path succeeds
`send.sticker` / `imsg send-sticker` fails on every attempt with:

```
Dylib error: Could not securely open sticker directory
```

This happens even though `imsg status --json` reports `selectors.stickerSend: true` and `send.sticker` in `rpc_methods` — so the capability probe advertises support and the send is then refused unconditionally.

**Environment:** macOS 26, imsg 0.13.4, SIP disabled, dylib injected via `imsg launch` (confirmed with `lsof` that Messages has `0.13.4/libexec/imsg-bridge-helper.dylib` loaded, not a stale one).

## The rich-link path is the control

Both features stage into `~/Library/Messages/Attachments/imsg/…` and both guard the staging directory with a "secure open" helper. On the same machine, in the same session, **rich links work and stickers don't** — verified in chat.db, where a bare-URL send produced `balloon_bundle_id = com.apple.messages.URLBalloonProvider` with a 1193-byte `payload_data`, versus the same URL sent before the feature was enabled which had neither.

The two helpers differ in exactly one way that matters.

`openRichLinkDirectorySecurely` (`IMsgInjected.m`) opens the **trusted root itself** and walks downward from there:

```objc
NSString *root = trustedRichLinkStagingRoot();   // ~/Library/Messages/Attachments/imsg
int directoryFD = open(root.fileSystemRepresentation, O_RDONLY | O_CLOEXEC | O_DIRECTORY);
```

`openUserOwnedDirectorySecurely`, used by `openStickerDirectorySecurely`, starts at the **user's home directory** and walks down through every component:

```objc
NSString *home = [actualUserHomeDirectory() stringByStandardizingPath];   // /Users/<user>
int directoryFD = open(home.fileSystemRepresentation,
                       O_RDONLY | O_CLOEXEC | O_DIRECTORY | O_NOFOLLOW);
if (directoryFD < 0) return -1;
```

## Why that first `open` cannot succeed

Messages.app is sandboxed (`com.apple.security.app-sandbox`), and injecting a dylib doesn't lift the sandbox. Its file-access exceptions are narrow:

```
com.apple.security.temporary-exception.files.home-relative-path.read-write
    → /Library/Messages/
com.apple.security.exception.files.home-relative-path.read-only
    → /Library/com.apple.ManagedSettings/EffectiveSettings.plist
```

So `~/Library/Messages/…` is reachable, but the home directory itself is not in any granted path. `open("/Users/<user>", O_RDONLY | O_DIRECTORY)` is denied, the helper returns -1 before it ever looks at the staging directory, and the caller reports the generic "could not securely open sticker directory".

That also explains the asymmetry precisely: the rich-link helper's first syscall lands *inside* the granted subtree, the sticker helper's first syscall lands *outside* it. Same intent, different starting point, opposite outcomes.

## Ruling out the obvious alternatives

- **Not permissions.** Every component of the path is user-owned and non-world-writable: `/Users/<user>` `drwxr-x---`, `Library` `drwx------`, `Messages` `drwxrwxr-x`, `Attachments` `drwxr-xr-x`, `imsg` `drwxr-xr-x`, `stickers` `drwx------`. (`~/Library/Messages` started out `drwxrwxrwx`, which would have tripped the `S_IWOTH` check; fixing it changed nothing.)
- **Not a symlink.** No component is a symlink, so `O_NOFOLLOW` isn't the trigger.
- **Not the source path.** The failure is identical whether `--file` points at an arbitrary directory or at a file already copied into `~/Library/Messages/Attachments/imsg/stickers/`.
- **Not staging.** The staging copy succeeds — `imsg/stickers/` gets created on the first attempt. The refusal comes afterwards, when the dylib tries to open what was staged.
- **Not a stale bridge.** Reproduced after a full quit-and-relaunch of Messages with the current dylib. (Worth noting separately: `imsg launch` prints "Messages.app launched with dylib injection" even when `ensureRunning()` no-ops because an older resident bridge still answers ping — that cost some time before I checked `lsof`. Possibly worth a distinct message for the no-op case.)

## Suggested fix

Give the sticker path the same shape as the rich-link path: open `trustedStickerRoot()` (or `trustedMessagesAttachmentsRoot()`) directly and walk down from there, rather than starting at the home directory. The home-directory hop doesn't add a meaningful guarantee — the trusted root is a fixed absolute path under `~/Library/Messages/`, which is exactly the subtree the sandbox already grants — and it's the only reason the two paths behave differently.

Happy to test a patch; the reproduction here is deterministic.

steipete:
## Summary

- anchor secure sticker and transfer directory traversal at the fixed trusted Messages attachments root
- retain descriptor ownership, world-write, directory, and no-symlink checks
- update source-level regression coverage and the 0.14.0 changelog

Closes #211.

## Verification

- `make test` — 510 tests passed
- `make lint` — passed with 12 existing non-serious warnings
- `make build-dylib` — built the real injected helper successfully
- `swift package update --dry-run` — zero dependency changes
- built CLI: version/help/status/invalid sticker invocation exercised; status correctly reports SIP enabled, so a sandboxed Messages send cannot run on this host
- Codex autoreview — clean, no accepted/actionable findings


steipete:
Landed as ef8e2aef35becc31101f9f4f90f4336db1234790.

Verification:

- `make test`: 510 tests passed.
- `make lint`: passed; 12 existing non-serious warnings only.
- `make build-dylib`: compiled the real injected helper successfully.
- `swift package update --dry-run`: zero dependency changes.
- Built CLI behavior: `imsg --version`, `imsg send-sticker --help`, `imsg status --json`, and an invalid sticker invocation were exercised. The CLI correctly reports that this host has SIP enabled, so live Messages dylib injection and an actual sticker delivery were not possible here.
- Exact-head CI run 31291977083: macOS and Linux read-core jobs passed for dd73f692ba881e60c195f716f58a89f01192a94f.
- Codex autoreview: clean, no accepted/actionable findings.

The remaining live sandbox-send confirmation requires a SIP-disabled Mac with the helper injected; the source-level regression, full suite, real helper build, and exact-head CI are all green.

PR #214, opened 2026-08-09
fix(rpc): reject malformed requests

steipete:
## Summary

- make every `imsg rpc` method validate named parameters through one strict decoder
- reject unknown keys, type coercions, alias conflicts, and conflicting chat targets before reads or side effects
- enforce JSON-RPC 2.0 request and notification semantics with actionable error codes
- preserve the poll selector and attachment reply behavior landed in #213

## Problem

Several older RPC handlers treated malformed values as absent. A misspelled or wrongly typed `watch.subscribe` filter could therefore become an all-chat subscription, while multiple chat selectors allowed `chat_id` to silently override a different identifier or GUID. Other invalid caller inputs were reported as internal failures, and failed notifications incorrectly emitted response envelopes.

## Solution

The RPC layer now has one typed parameter boundary with explicit method key sets and documented aliases. Strings, integers, booleans, arrays, and IDs are no longer coerced. Methods accepting chat targets require exactly one selector, and direct recipients remain mutually exclusive with chat selectors. Caller validation happens before cache access, staging, bridge invocation, or mutation.

## Compatibility decision

This intentionally stops accepting RPC input that did not conform to the documented JSON-RPC contract, including omitted protocol versions, positional params, scalar coercions, ignored unknown keys, and multiple simultaneous chat selectors. We are accepting that upgrade break rather than adding a compatibility mode because the permissive behavior could silently broaden message subscriptions or redirect mutations to a different chat. Documented request shapes and the aliases listed in the RPC reference remain supported.

## Validation

- `make lint`
- `make test` — 527 tests across 5 suites
- `make build` — universal signed CLI and bridge helper
- `git diff --check`
- Codex autoreview — no accepted/actionable findings

## Live proof

The universal candidate was copied to an owner-only temporary directory on a private macOS test host and run against a real Messages database. The existing private bridge reported v2 ready with advanced features, typing indicators, and read receipts available.

A valid `chats.list` request succeeded. Eight malformed request classes returned `-32602` before subscription or mutation: wrong-type `chat_id`, `chatId` typo, mixed-type participants, string boolean, array params, unknown key, conflicting destructive chat selectors, and recipient plus chat selector. Two invalid notifications emitted no output. No message or chat mutation was performed, and all temporary files were removed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@
 ### JSON-RPC
 - fix: match CLI poll option resolution and preserve reply part indexes for attachment sends.
 - feat: add bounded `messages.after` pagination with authoritative database-instance-scoped ROWID cursors, cross-chat catchup, and optional standalone reaction events (#200, #201, thanks @vincentkoc).
-- fix: validate JSON-RPC framing, IDs, named params, types, aliases, and chat-target conflicts consistently so malformed requests cannot broaden or redirect operations.
 
 ### Advanced IMCore
 - fix: open trusted sticker staging roots directly so sandboxed Messages can send staged stickers (#211, thanks @clawcrab).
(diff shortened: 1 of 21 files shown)


steipete:
Addressed the review: the release-owned CHANGELOG entry was removed, and the PR body now explicitly accepts the intentional RPC compatibility break. Code is unchanged; the one-line removal had clean autoreview.

PR #217, opened 2026-08-10
feat(rpc): add degraded startup status

steipete:
## What Problem This Solves

`imsg rpc` previously could not stay available when `chat.db` was missing or unreadable. Startup either exited or exposed only a one-shot startup error, so clients had no in-band way to discover capabilities or recover after database access became available. Schema probes could also swallow SQLite failures and incorrectly treat a sidecar-less WAL snapshot as ready.

Bridge readiness had a similar ownership problem: the existence of the v2 inbox was conflated with registry and action-selector readiness, while stale probes and cancellation produced misleading observable outcomes.

## Why This Change Was Made

The RPC process now owns a degraded-but-live control plane. Protocol v1 `initialize` and `status` report database, bridge, contacts, supported methods, and the methods usable at that moment. Database resources retry in-band, schema inspection fails closed, and unavailable resources return typed JSON-RPC errors: `-32002` for database unavailability and `-32003` for bridge unavailability.

Bridge capability reporting now separates v2 inbox readiness, chat-registry readiness, and individual action selectors. Status and capability probes, along with other bridge-only methods, use a supervised, non-launching bridge path: they do not launch, kill, or relaunch Messages. Released `typing` compatibility is preserved through its TypingIndicator fallback, and `read` compatibility is preserved through IMCoreBridge activation; either may activate Messages. Direct AppleScript `send` remains allowed to activate Messages, as documented.

## User Impact

Supervisors can start `imsg rpc` before Full Disk Access or a usable database is present, inspect the exact dynamic method set, repair access or replace the database at the configured path, and recover without restarting the RPC child. Clients can negotiate protocol v1 and distinguish unavailable dependencies from malformed requests or internal failures.

## Evidence

- Full test suite: 596 passed.
- Focused readiness, capability, and client tests: 31 passed after the final test split.
- RPC-focused tests: 151 passed; focused typing/read tests: 10 passed.
- `make lint` passed, with new test files below 500 lines and no new warning.
- Universal `make build` passed: CLI arm64+x86_64; helper arm64e+arm64+x86_64.
- `git diff --check` passed.
- Fresh autoreview passed with no actionable findings (0.94 confidence).
- Live macOS same-child recovery proof covered initial degraded status, fail-closed handling of a sidecar-less WAL backup, recovery after replacing the database at the same configured path, successful `chats.list`, protocol v1 initialization, non-launching bridge status, clean EOF, and empty stderr.
- CI remains explicit: the macOS workflow runs on `macos-26` and records `sw_vers`; Linux read-core remains on `ubuntu-latest` with Swift 6.3.3.


steipete:
@clawsweeper re-review

steipete:
Both ClawSweeper findings are addressed: `typing` is restored to its TypingIndicator fallback, and `read` is restored to IMCoreBridge activation. Direct-recipient and GUID targets work without the database; `chat_id` remains database-required.

Focused proof: 151 RPC tests passed, 10 typing/read tests passed, `make lint` passed, and `git diff --check` passed. Fresh autoreview is clean with no actionable findings (0.94 confidence).

@clawsweeper re-review

PR #219, opened 2026-08-10
feat(rpc): complete message contract parity

steipete:
## What Problem This Solves

The RPC surface had drifted from the canonical CLI message contract. Chat payloads were assembled separately, database search was unavailable over RPC, rich sends did not accept the CLI file/path form, multipart text sends were absent, method metadata and dispatch could diverge, and several semantic validations were weaker than their documented contracts.

More importantly, newly exposed rich-file and multipart sends could appear successful when the bridge helper acknowledged publication but no outgoing message row was observable. That false acknowledgement is unsafe because callers may treat the send as complete.

## Why This Change Was Made

This change makes the CLI payload models and one descriptor-owned registry authoritative across method discovery, capability reporting, request lanes, and dispatch. It adds bounded database-backed message search, text-only multipart sends, rich file/path parity, positive limit and poll validation, the narrow `allow_sms_fallback` control, and stable attachment field documentation.

Rich-file and multipart sends now cross a stricter post-publication boundary: success requires observing the matching outgoing database row. If publication may have occurred but the row cannot be observed, the RPC returns typed `-32001` delivery-outcome-unknown data with `may_have_completed`, never `ok`. Callers must not automatically retry that result.

The production delta is approximately +653/-245 lines, with +537/-15 lines of tests and +56/-16 lines of documentation. The growth is additive public protocol capability plus explicit ownership of delivery-safety verification.

## User Impact

RPC clients can use `messages.search`, text-only `send.multipart`, rich `file`/`path` sends, and `allow_sms_fallback`. Chat objects now share the CLI's canonical payload owner, method support and dispatch are derived from the same descriptors, positive bounds are enforced consistently, and attachment keys are documented as `total_bytes` and `original_path`.

For new rich-file and multipart sends, an unobserved post-publication result is deliberately surfaced as uncertain rather than successful. This may expose previously hidden delivery uncertainty, but prevents false success and unsafe retries.

## Evidence

- Full macOS test suite: 623 passed.
- `make lint`: green, with only prior nonfatal warnings.
- Universal build: green.
- Linux tests and build: green.
- Diff/whitespace check: clean.
- Final structured autoreview: clean, 0.97.
- Live read-only validation: supported methods include search, multipart, and rich sends; multipart is currently usable; bounded zero-result search works; CLI and RPC first-chat keysets match at 13 keys each; RPC framing is clean.
- Live mutation validation made exactly one multipart request and one rich-file request. Each helper returned success, but neither produced an observable outgoing database row. Neither request was retried. These attempts do not prove successful delivery; they exposed the false acknowledgement that the strict observation requirement now prevents.
- Post-recovery production topology was restored and is healthy: the retired gateway is disabled, the obsolete RPC count is zero, one current production-owned RPC and one Messages process remain, and v2 reports ready.
- Regression coverage includes observed, nil, error, unavailable-database, notification, empty-caption, and single-publication outcomes. It proves uncertain rich-file/multipart publication returns typed `-32001` with `may_have_completed` and never `ok`.

Required CI gates are `macos` on the `macos-26` runner and `linux-read-core`.

### DIFF
--- a/Sources/IMsgHelper/IMsgInjected.m
+++ b/Sources/IMsgHelper/IMsgInjected.m
@@ -1085,6 +1085,7 @@ static NSDictionary* handleStatus(NSInteger requestId, NSDictionary *params) {
         @"sendMessageReason": @(gHasSendMessageReason),
         @"sendMessage": @(chatSend),
         @"sendAttachment": @(attachmentSend),
+        @"sendMultipart": @(chatSend),
         @"sendReaction": @(reactionSend),
         @"pollPayloadMessage": @(
             [chatClass instancesRespondToSelector:@selector(sendMessage:)]
--- a/Sources/imsg/Commands/ChatsCommand.swift
+++ b/Sources/imsg/Commands/ChatsCommand.swift
@@ -77,21 +77,4 @@ enum ChatsCommand {
     }
   }
 
-  private static func contactNameForChat(
-    chat: Chat,
-    chatInfo: ChatInfo?,
-    participants: [String],
-    contacts: any ContactResolving
-  ) -> String? {
-    let identifier = chatInfo?.identifier ?? chat.identifier
-    let guid = chatInfo?.guid ?? ""
-    guard !isGroupHandle(identifier: identifier, guid: guid) else { return nil }
-    if let name = contacts.displayName(for: identifier) {
-      return name
-    }
-    if participants.count == 1 {
-      return contacts.displayName(for: participants[0])
-    }
-    return nil
-  }
 }
--- a/Sources/imsg/Commands/SendRichCommand+Support.swift
+++ b/Sources/imsg/Commands/SendRichCommand+Support.swift
@@ -46,17 +46,61 @@ extension SendRichCommand {
         Int64?,
         Date
       ) async throws -> Message?,
-    storeFactory: (String) throws -> MessageStore
+    storeFactory: (String) throws -> MessageStore,
+    resolveRow: Bool = false
   ) async throws -> [String: Any] {
+    let store = try? storeFactory(dbPath)
+    let chatInfo = try? store?.chatInfo(matchingTarget: chat)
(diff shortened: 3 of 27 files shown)


steipete:
Exact-head proof for `3ebdb445412ddfaf791ca8d117ecbdd7e49ac660`:

- CI attempt 1 had an isolated macOS `make test` runner hang for 39 minutes and was cancelled after the focused local tests completed with no lingering tasks. The same immutable run's attempt 2 succeeded: macOS 26 Test in 80s, Build in 60s, and `linux-read-core` green.
- Local proof is green: 623 tests, universal build, lint, Linux tests/build, clean diff/whitespace, and final structured autoreview clean at 0.97.
- Live validation made exactly one multipart request and one rich-file request. Both helpers reported success without an observable outgoing row; neither request was retried. That false-success discovery is the reason automatic retry is forbidden.
- The strict verifier now returns typed `-32001` delivery-outcome-unknown data with `may_have_completed` and never `ok` when post-publication observation is unavailable or unsuccessful; focused regressions cover observed, nil, error, unavailable-database, notification, empty-caption, and single-publication outcomes.

@clawsweeper re-review

steipete:
Resolved both durable P1 findings on the new head:

- `send.multipart` now requires database readiness, and both rich-file and multipart sends acquire the database plus a maximum-ROWID baseline before any publication.
- Strict success now requires a nonempty helper `messageGuid` that resolves above that baseline in the target chat. Missing, old, wrong-chat, or lookup-error GUIDs return `may_have_completed`; an unavailable database fails pre-dispatch with `-32002` and does not call the bridge.

Proof: focused status/send/delivery tests 36/36; RPCSendParity stress 50/50; full `make test` 624/624; lint and diff checks clean; fresh structured review found no actionable P0/P1 findings.

@clawsweeper please re-review the new head.

steipete:
@clawsweeper re-review

PR #224, opened 2026-08-10
build: track SwiftPM release lock

steipete:
## What Problem This Solves

Release run [31401419580](https://github.com/openclaw/imsg/actions/runs/31401419580) passed credential validation on attempt 2 but stopped before tagging because the shared Swift release workflow requires a tracked `Package.resolved`. The repository generated a valid lock locally while `.gitignore` explicitly excluded it.

No tag, draft release, artifact, or Homebrew update was created by the failed attempts.

## Why This Change Was Made

Track the existing SwiftPM lock and stop ignoring it so signed release builds resolve the exact reviewed dependency revisions. A packaging regression test now requires the lock to exist in a clean checkout and remain outside `.gitignore`.

## User Impact

This unblocks the frozen imsg 0.14.0 release and makes future signed builds reproducible across macOS and Linux release jobs.

## Evidence

- `swift package resolve` left `Package.resolved` byte-for-byte unchanged
- release packaging tests: 8 passed
- full suite: 653 passed
- touched-file Swift format and SwiftLint: clean, 0 violations
- `git diff --check`
- independent autoreview: clean, no accepted/actionable findings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,7 +34,6 @@ go.work.sum
 # SwiftPM
 .build/
 .swiftpm/
-Package.resolved
 *.xcodeproj
 
 # Build artifacts
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,0 +1,33 @@
+{
+  "originHash" : "3b4b8fcc219563b7f4bea10a00c2ede7c4c4bd385bfd1cf2337a20a1256b3180",
+  "pins" : [
+    {
+      "identity" : "commander",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/steipete/Commander.git",
+      "state" : {
+        "revision" : "bd219c4ee9032fee3e009856f81fcc6ec09a85f4",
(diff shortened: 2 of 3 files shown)


steipete:
Land-ready on `236f634606758a46ebfcb0f7d35278f86f98517f`.

The shared release workflow’s clean-checkout preflight cannot be exercised against this patch until it reaches protected `main`; that is the next release step, not an unaddressed code finding. The equivalent repository-owned proof is complete:

- exact-head CI: macOS and Linux succeeded
- `swift package resolve` left the committed lock byte-for-byte unchanged
- packaging tests: 8 passed
- full suite: 653 passed
- touched-file lint: 0 violations
- `git diff --check` passed
- ClawSweeper: no actionable findings
- final full-branch independent review: clean

After merge, I will dispatch the same 0.14.0 release workflow anew against the new immutable `main` commit and require its preflight to proceed before publication.

PR #235, opened 2026-08-11
feat(rpc): add caller-owned tracked send IDs

### LINKED ISSUE #234: RPC: add caller-owned GUIDs for exact send-timeout reconciliation
## Problem

`imsg` now correctly fails closed after an uncertain transport outcome, but a
remote RPC caller can still lose the response after Messages accepted a send.
The caller then knows that delivery *may* have happened, but it has no exact
message identity to query. Matching history by recipient, text, and time can
misidentify an unrelated identical send.

## Proposed additive contract

Add a narrow `send.tracked` JSON-RPC method:

- require a caller-supplied UUID `attempt_id`;
- support exactly one text message through the already-running IMCore bridge;
- use `attempt_id` as the outgoing `IMMessage` GUID;
- return the same value as `attempt_id`, `guid`, and `message_id`;
- never use AppleScript fallback and never retry after uncertainty;
- advertise the method only when the running injected helper reports that it
  can preserve a caller-owned message GUID.

After a lost response, the caller can query the existing
`message.send_status` method with that exact UUID. This is ownership evidence,
not a general idempotency or duplicate-suppression promise.

The existing `send` method remains unchanged for compatibility. Attachments
and multi-message/chunked sends stay outside the tracked contract so one marker
always identifies one platform message.

## Acceptance evidence

- RPC tests for UUID validation, stale-helper rejection before dispatch,
  capability advertisement, exact helper forwarding, and exact response IDs.
- Static helper tests covering UUID validation and all supported IMMessage
  constructor GUID slots.
- Hosted compile/test checks.
- Follow-up Mac proof that the supplied UUID persists into `chat.db` and is
  returned by `message.send_status` before downstream adoption.

clawSean:
## Summary

- add an additive `send.tracked` RPC method for one bridge-backed text message
- require a caller UUID and preserve it as the outgoing `IMMessage` GUID
- fail closed before dispatch on invalid IDs, attachments, AppleScript, or a stale helper
- advertise the method only when the running helper proves caller-owned GUID support
- return the same exact ID as `attempt_id`, `guid`, and `message_id`
- document timeout reconciliation through the existing `message.send_status` method

The existing `send` method and its fallback behavior are unchanged.

Fixes #234.

## Why

Transport-level uncertainty handling prevents unsafe automatic retries, but a
remote caller can still lose the response after Messages accepted a send. A
caller-owned GUID lets that caller query one exact attempt instead of matching
history by recipient, text, and time, which can falsely identify another
identical send.

This is an ownership marker, not a general idempotency promise. The contract is
deliberately one marker to one platform message.

## Validation

- `git diff --check`
- added focused RPC tests for UUID validation, stale-helper rejection before
  dispatch, exact bridge forwarding, and exact response IDs
- added status-capability tests so stale helpers never advertise the method
- added static helper coverage for UUID validation and every supported
  `IMMessage` GUID constructor slot

The development host is a 2-vCPU Linux VPS without Swift. The trusted Mac proof
node is currently offline, so Swift compile/tests and helper build are left to
hosted CI. A live Mac `chat.db` + `message.send_status` proof should be completed
before downstream adoption.


steipete:
Maintainer lane review completed at `c61f2ac8ca85097f15a759fb66b30e7192617446`.

I preserved the contributor commit and added a focused hardening commit. The original implementation allowed a caller-selected UUID to collide with an existing message or a concurrent caller's tracked send. The branch now:

- rejects case-insensitive GUID collisions from message history before bridge publication;
- atomically reserves tracked GUIDs in the injected helper shared by all RPC children;
- advertises and accepts `send.tracked` only when the helper reports the stronger reservation capability, so stale helpers fail closed;
- carries authoritative helper duplicate rejection back as `not_started` / `retry_safe: true`;
- retains `may_have_completed` and `still_in_flight` as non-retryable and never falls back to AppleScript.

Proof:

- `make test`: 278 tests passed.
- `make lint`: exit 0 (warning-only existing size findings).
- `make build`: universal `imsg` and helper artifacts built and signed locally.
- Real built `imsg rpc` process: malformed UUID, attachment, and forced-AppleScript tracked requests each returned one `-32602` response; no success/send response was emitted.
- Hosted CI run `31565233192`: `macos` and `linux-read-core` succeeded; PR checks are green and merge state is clean.
- Final structured branch autoreview: clean, no accepted/actionable findings.

No live message was sent; the lane explicitly required deterministic fault injection only. The branch is ready for orchestrator review and landing. I did not merge it.

